PR #16, opened 2023-06-07
:sparkles: Add text generation module

gkumbhat:
### Changes

- Add text generation module to support inference from Seq2Seq and CausalLM models

NOTE: this currently only supports TGIS backend.


gabe-l-hart (commented):
A couple of little NITs, but nothing blocking. Needs to get `lint` passing too

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+        with saver:
+            artifacts_dir = "artifacts"
+            saver.update_config(
+                {
+                    "artifact_path": artifacts_dir,
+                    "bos_token": self._bos_token,
+                    "sep_token": self._sep_token,
+                    "eos_token": self._eos_token,
+                    # "truncate_input_tokens": self._truncate_input_tokens,
```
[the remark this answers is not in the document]
gkumbhat: removed.

caikit_nlp/resources/pretrained_model/base.py
```diff
@@ -209,16 +209,16 @@ def load(cls, model_path: str) -> Type["PretrainedModelBase"]:
         )
         return res
 
-    def save(self, model_path: str):
+    def save(
+        self, model_path: str, tokenizer_dirname: str = "", base_model_dirname: str = ""
```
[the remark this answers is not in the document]
gkumbhat: I think it shouldn't break anything, since this `save` function isn't getting used directly currently and since `load` function of this base class isn't using those dirnames directly, it won't break loading of these saved model as well. 

Idea behind this change was to let user of this module direct how the artifacts can be saved and changed the defaults, since the usual convention with the transformer models is to save both tokenizer and models in same directory. Having them saved as `""` would make the models we save with this function directly loadable via HF load function as well (since there won't be a separate path inside the folder).

caikit_nlp/modules/text_generation/__init__.py
```diff
@@ -15,3 +15,4 @@
 # Local
 from .peft_prompt_tuning import PeftPromptTuning
 from .peft_tgis_remote import PeftPromptTuningTGIS
```
alex-jw-brooks, line 17: Metacomment - in an unrelated PR, should we make the TGIS client? It's a bit awkward since a model does need to be bootstrapped, but I do feel like either this or the Peft prompt tuning block could be used to initialize it

tests/modules/test_text_generation.py
```diff
@@ -0,0 +1,77 @@
...
+
+### Helper stubs / mocks; we use these to patch caikit so that we don't actually
+# test the TGIS backend directly, and instead stub the client and inspect the
+# args that we pass to it.
+class StubClient:
+    def __init__(self, base_model_name):
+        pass
+
+    # Generation calls on this class are a mock that explodes when invoked
+    Generate = mock.Mock(side_effect=RuntimeError("TGIS client is a mock!"))
```
alex-jw-brooks, line 29: Can we add the stub client/stub backend somewhere common to avoid duplicating it across this and the TGIS remote tests?

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -0,0 +1,261 @@
...
+        get a caikit text-generation 'model'.
+
+        Args:
+            base_model_path: str
+                Path to transformers model
+                NOTE: Model path needs to contain tokenizer as well
+            load_backend: BackendBase
+                Backend object to be used to run inference with.
+                NOTE: this is required for inferencing. It is
+                made optional just in provide support for model conversion use-case
```
alex-jw-brooks, line 110:
```suggestion
                made optional to support the model conversion use-case
```
By model conversion, you mean to ORT etc, right? Could you elaborate a bit more on what this looks like?

I had was under the impression that TGIS model conversions generally happen in memory, and there's not an export API, so I don't think I understand how you'd leverage that without inference support

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -0,0 +1,261 @@
...
+                    model_config.model_type
+                ),
+            )
+        log.debug("Bootstrapping base resource [%s]", base_model_path)
+        base_model = resource_type.bootstrap(
+            base_model_path, tokenizer_name=base_model_path
+        )
+        bos_token = base_model._tokenizer.bos_token
+        sep_token = base_model._tokenizer.sep_token
+        eos_token = base_model._tokenizer.eos_token or None
```
alex-jw-brooks, line 137: Why does only EOS have a fallback here? Shouldn't any fallback behavior be handled in bootstrap of the base model?

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -0,0 +1,261 @@
...
+            model_path=model_path,
+        )
+        with saver:
+            artifacts_dir = "artifacts"
+            saver.update_config(
+                {
+                    "artifact_path": artifacts_dir,
+                    "bos_token": self._bos_token,
+                    "sep_token": self._sep_token,
+                    "eos_token": self._eos_token,
```
alex-jw-brooks, line 167: Pad token should also be exportable and reloaded I think

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -0,0 +1,261 @@
...
+        Returns:
+            TextGeneration
+                Instance of this class built from the on disk model.
+        """
+        error.type_check("<FPT03521359E>", TGISBackend, load_backend=load_backend)
+
+        config = ModuleConfig.load(model_path)
+        base_model_path = config.get("artifact_path", "")
+        base_model_path = os.path.join(model_path, base_model_path)
+        error.dir_check("<DWC20623231E>", base_model_path)
```
alex-jw-brooks, line 196: Inconsistent log code prefix

alex-jw-brooks (changes requested):

caikit_nlp/modules/text_generation/__init__.py
gkumbhat: this is related to `.peft_tgis_remote` module right? Yeah, I think that make sense.

caikit_nlp/modules/text_generation/text_generation.py
gkumbhat: The model conversion here refers to both converting the model if required by TGIS (like ORT) or just converting the model into `caikit` format, since at load time this module (in runtime) will only support loading of a `caikit` model and not download / load models that are in `huggingface` format

caikit_nlp/modules/text_generation/text_generation.py
gkumbhat: this is just to assign eos_token to something that gets ignored and the decision of "ending the sequence" is then left to the model.

caikit_nlp/modules/text_generation/text_generation.py
gkumbhat: yep. good point. Will add that here

caikit_nlp/modules/text_generation/text_generation.py
gkumbhat: 🤦

gabe-l-hart:
Not blocking for this PR, but see https://github.com/caikit/caikit/pull/237 for my plan on the centralized interfaces for Text Generation

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -0,0 +1,262 @@
...
+
+        config = ModuleConfig.load(model_path)
+        base_model_path = config.get("artifact_path", "")
+        base_model_path = os.path.join(model_path, base_model_path)
+        error.dir_check("<FPT01983374E>", base_model_path)
+        return cls(
+            base_model_path,
+            bos_token=config.bos_token,
+            sep_token=config.sep_token,
+            eos_token=config.eos_token,
```
alex-jw-brooks, line 202: pad token should also be passed here!

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -0,0 +1,262 @@
...
+            tgis_backend=load_backend,
+        )
+
+    def run(self, text, preserve_input_text=False):
+        """Run inference against the model running in TGIS.
+
+        Args:
+            text: str
+                Source string to be encoded for generation.
+            preserve_input_text: str
```
alex-jw-brooks, line 212:
```suggestion
            preserve_input_text: bool
```

alex-jw-brooks (approved):
LGTM! thanks, this is great

PR #17, opened 2023-06-07
Update caikit

gkumbhat:
### Changes
- Update caikit to `0.7.0`
- Update caikit-tgis-backend to `0.1.6` as that can work with `caikit 0.7.0`
- Rename `blocks` to `modules` as per new caikit paradigms
- Add `text-generation-task`


caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -11,9 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""This module contains a block for prompt tuning through PEFT. Currently,
+"""This module contains prompt tuning through PEFT. Currently,
 we are using this to validate our abstractions and design additional necessary
```
alex-jw-brooks, line 15: Meta note - we should remove this 😄

caikit_nlp/modules/text_generation/text_generation_task.py
```diff
@@ -0,0 +1,27 @@
...
+
+# Local
+from ...data_model import GeneratedResult
+
+
+@task(
+    required_parameters={"text": str},
+    output_type=GeneratedResult,
+)
+class TextGenerationTask(TaskBase):
```
alex-jw-brooks, line 27: I need to catch up with caikit task changes, but what information (if anything, assuming @task is basically the secret sauce with parameter signatures etc) is meant to be implemented on this class?

I.e., is this supposed to replace the abstract module type base class?

caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -40,24 +40,34 @@
...
+            self._client = tgis_backend.get_client(base_model_name)
+
         self.base_model_name = base_model_name
         self._prompt_cache_id = prompt_cache_id
         self.eos_token = eos_token
         self.verbalizer = verbalizer
 
     @classmethod
-    def load(cls, model_path: str) -> "PeftPromptTuningTGIS":
+    def load(cls, model_path: str, load_backend: BackendBase) -> "PeftPromptTuningTGIS":
```
alex-jw-brooks, line 70: Given that this backend is TGIS specific, do we need additional validation here? I'm wondering what the advantage of parameterizing this are

tests/modules/foo_test_text_generation.py
```diff
@@ -0,0 +1,48 @@
...
+    """Check if we can bootstrap and run causallm models"""
+    # Patch our stub backend into caikit so that we don't actually try to start TGIS
+    mock_get_backend.return_value = StubBackend()
+
+    model = TextGeneration.bootstrap(CAUSAL_LM_MODEL)
+
+    sample_text = "Hello stub"
+    response = model.run(sample_text, preserve_input_text=True)
+
+    assert isinstance(response, GeneratedResult)
```
alex-jw-brooks: Why is this file name prefixed with `foo` to prevent pytest from grabbing it. Should we remove it?

alex-jw-brooks (changes requested):
Thanks, this looks great! Some thoughts and my questions, mostly just on the intended patterns with this change

caikit_nlp/modules/text_generation/text_generation_task.py
gkumbhat: Honestly I am also yet to dig deeper into task logic to understand tasks more. I have the same question.

> is this supposed to replace the abstract module type base class?

What to you mean by this?

caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -40,24 +40,34 @@
...
+            verbalizer,
+            enable_backend=True,
+            tgis_backend=None,
+        ) -> None:
         super().__init__()
         # Configure the internal client
-        self._client = module_backend_config.get_backend(
-            TGISBackend.backend_type
-        ).get_client(base_model_name)
+        if enable_backend:
```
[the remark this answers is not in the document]
gkumbhat: This is actually a workaround paradigm we are following until we decide on how "shared backend" stuff is going to play out finally in caikit.

tests/modules/foo_test_text_generation.py
gkumbhat: sorry, this was accidentally pushed. I have removed it in my latest commit

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: done

caikit_nlp/modules/text_generation/peft_tgis_remote.py
alex-jw-brooks: I see - but will those ever be instances that aren't `TGISBackend`? I am mostly wondering if we should throw a TypeError if a different implementation of the BackendBase is passed (except for testing via the stub stuff maybe)

caikit_nlp/modules/text_generation/text_generation_task.py
alex-jw-brooks: Is this meant to replace what would be `TextGenerationBase`, or is it completely separate?

Since we never got around to adding it and still inherit from `ModuleBase` directly

caikit_nlp/modules/text_generation/peft_tgis_remote.py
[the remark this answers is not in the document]
alex-jw-brooks: In what way though? Because if this is `False`, .run() fails due to an uninitialized client, right?

caikit_nlp/modules/text_generation/text_generation_task.py
gkumbhat: This is completely separate. It needs to be there for every "type" of module and essentially defines the shape of input / output for a given "task"

caikit_nlp/modules/text_generation/peft_tgis_remote.py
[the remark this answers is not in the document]
gkumbhat: yes, `enable_backend` is essentially meant to handle the case where we want to initialize the class but not run inference, for example, in case of model-bootstrap+conversion. Actually I added a note for this in other PR, I'll pull that over to here as well.

Also, I'll add a check with relevant error message in case this is empty and someone tries to use it for inferencing.

caikit_nlp/modules/text_generation/peft_tgis_remote.py
gkumbhat: Likely not. I'll add error check

caikit_nlp/modules/text_generation/text_generation_task.py
evaline-ju: Tasks allow for the verification of correct run signatures. There can be different modules with the same task.

alex-jw-brooks: Cool sounds good, just wanted to better understand if the content of the task is largely the decorator, or the decorator + things to be added to the class it's wrapping. Thanks!

alex-jw-brooks (approved):
Looks great! LGTM 😄

Issue #29, opened 2023-06-09
Fix logic for validating output model types


kellyaa:
## Describe the bug

Current logic has an error in validating `output_model_types`. This is a list, and the logic is looking for a list in a list, but should be comparing the intersection of the lists.
https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L333
```tuning_config.output_model_types in base_model.PROMPT_OUTPUT_TYPES,```

Also, there is a reference to `base_model.model_type` which should be `base_model.MODEL_TYPE`

kellyaa:
PR merged!

PR #30, opened 2023-06-09
Fix logic for model type checking

kellyaa:
Addresses #29

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
+++ b/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
@@ -415,14 +415,27 @@ class PeftPromptTuning(ModuleBase):
         if not tuning_config.output_model_types:
             output_model_types = base_model.PROMPT_OUTPUT_TYPES
         else:
+            # If the first element is not PromptOutputModelType, assume the entire list isn't and convert
+            if not isinstance(
+                tuning_config.output_model_types[0], PromptOutputModelType
+            ):
+                output_model_types = []
+                for output_type in tuning_config.output_model_types:
+                    output_model_types.append(PromptOutputModelType(output_type))
+            else:
+                output_model_types = tuning_config.output_model_types
             error.value_check(
                 "<NLP36947542E>",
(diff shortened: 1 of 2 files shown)


caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -330,9 +330,15 @@ def train(
...
             error.value_check(
                 "<FPT36947542E>",
-                tuning_config.output_model_types in base_model.PROMPT_OUTPUT_TYPES,
+                len(
+                    list(
+                        set(tuning_config.output_model_types)
+                        & set(base_model.PROMPT_OUTPUT_TYPES)
+                    )
+                )
+                == len(tuning_config.output_model_types),
```
gkumbhat, line 339: Does this solution have issue if someone passes `output_model_types` as `List[str]` instead of `enum` ? since the `TuningConfig` datamodel accepts `List[str]` as data-type for `output_model_types`, I think we would need to explicitly convert it to enum for this validation and for further processing in this function.

gkumbhat (commented):
Thanks for contributing the fix @kellyaa . Can you also add some unit tests for this ?

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: I was thinking something like this:
```python
            # Change tuning config to PromptOutputModelType
            output_model_types = []
            for output_type in tuning_config.output_model_types:
                 output_model_types.append(PromptOutputModelType(output_type))

            error.value_check(
                "<FPT36947542E>",
                all(
                    [
                        output_type in base_model.PROMPT_OUTPUT_TYPES
                        for output_type in  output_model_types
                    ]
                ),
                "{} not supported for base model type {}".format(
                    tuning_config.output_model_types, base_model.MODEL_TYPE
                ),
            )

```

kellyaa:
Ready for review! FWIW these changes fixed my issues on the Ray side.

gkumbhat (approved):
LGTM. Thanks Kelly.

PR #34, opened 2023-06-14
Classification train record

Ssukriti:
The PR adds support for classification train record data model in tuning of text generation block . This is just to support multiple data formats while tuning. Further additions will be made to add classification module in future


gabe-l-hart (commented):
I haven't reviewed the PR in detail, but wanted to start a discussion here: I _think_ this kind of `data_model` work belongs in `caikit.interfaces.nlp` rather than in `caikit_nlp.data_model`. @alex-jw-brooks @gkumbhat @Ssukriti, where are you guys at w.r.t. this separation of interfaces from implementation?

caikit_nlp/data_model/classification.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+# Standard
+from typing import List
+
+# First Party
+from caikit.core import DataObjectBase
+
+# First party
+import alog
+import caikit
```
gkumbhat, line 24: nit: duplicate `First party` blocks for imports?

caikit_nlp/data_model/classification.py
```diff
@@ -0,0 +1,32 @@
...
+
+log = alog.use_channel("DATAM")
+
+
+@caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
+class ClassificationTrainRecord(DataObjectBase):
+    text: str
+    labels: List[str]
```
gkumbhat, line 32: should we expand this to `List[Union[str, int]]` ?  🤔

caikit_nlp/toolkit/task_specific_utils.py
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Local
+from ..data_model import ClassificationTrainRecord, GenerationTrainRecord
+
+
+def convert_to_generation_record(TrainRecord):
```
gkumbhat, line 19: nit: Can you please add docstring for this function explaining the intent

gkumbhat: since `TrainRecord` is a variable here, having it in camel case is a bit confusing. Can we change this to `snake_case` ?

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -360,6 +367,10 @@ def train(
             tuning_type = TuningType(tuning_type)
         error.type_check("<FPT65714993E>", TuningType, tuning_type=tuning_type)
 
+        train_stream = train_stream.map(convert_to_generation_record)
```
gkumbhat, line 370: should we add a type_check to make sure `train_stream` is of these 2 types? (same for val stream)

tests/modules/test_peft_prompt_tuning.py
```diff
@@ -125,6 +125,35 @@ def test_train_model(causal_lm_train_kwargs):
...
+    }
+    causal_lm_train_kwargs.update(patch_kwargs)
+    model = caikit_nlp.modules.text_generation.PeftPromptTuning.train(
+        **causal_lm_train_kwargs
+    )
+    # Test fallback to float32 behavior if this machine doesn't support bfloat16
+    assert model.model.dtype is torch.float32
+    # Ensure that we can get something out of it
+    pred = model.run("@bar what a cute cat!")
+    assert isinstance(pred, caikit_nlp.data_model.GeneratedResult)
```
gkumbhat, line 154: 🥳

gkumbhat (changes requested):
Thanks @Ssukriti for adding the classification train record support. Overall looks great. Just left some small nits.

caikit_nlp/data_model/classification.py
Ssukriti: sure but just mentioining internal conversion that will take place -

[label1,label2, label3] to "label1, label2, label3" because generation train record takes only string
then with int [1,2,3] it will be "1,2,3"

caikit_nlp/toolkit/task_specific_utils.py
Ssukriti: ya sorry, my bad

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
Ssukriti: handled this in        convert_to_generation_record function instead , if unsupported instance type while reading stream it will error

Ssukriti:
@gkumbhat I made the changes as requested, I will run the tests on a VM tomorrow and make sure they pass, then we can enable the workflows in fork here as well.
Feel free to take another look

Ssukriti:
@gkumbhat I tried adding union of List[str] or [int] to classification train record like this

`labels: Union[List[str], List[int]]`

but then in from proto and back / from json and back tests, it assigned labels as None . 

`def test_from_json_and_back():
        new = dm.ClassificationTrainRecord.from_json(classification_train_record.to_json())
        assert new.text == "It is 20 degrees today"
>       assert new.labels == ["temperature"]
E       AssertionError: assert None == ['temperature']
E        +  where None = ClassificationTrainRecord(text='It is 20 degrees today', labels=None).labels
`
2. Alternatively, if I try this List[Union[str, int]] I get error from py_to_proto `E  ValueError: Got unsupported entry type typing.Union[str, int]``

So I dont know how to support that Union in train record data model and need help with that.

gkumbhat (approved):
LGTM. Thanks Sukriti!

PR #39, opened 2023-06-19
:sparkles: Add embedding related data models

gkumbhat:


caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,188 @@
...
+    ]
+
+
+@dataobject(package="caikit_data_model.caikit_nlp")
+class Vector1D(DataObjectBase):
+    """Data representation for a 1 dimension vector.
+    NOTE: This only support float types
+    """
+
+    data: List[FloatValue]
```
gabe-l-hart, line 49: I think what you're trying to accomplish here is a representation of a consecutive sequence of numeric values backed by _some_ type of in-memory floating-point representation. As written, I think this is not the best way to represent that because as written, each element of the `List` can be a different in-memory type AND the `oneof` information needs to be held for each element of the list. I think a cleaner way to do this would be something like:

```py
@dataobject(...)
class Vector1D(DataObjectBase):

    @dataobject
    PyFloatSequence(DataObjectBase):
        values: List[float]

    @dataobject
    NpFloat32Sequence(DataObjectBase):
        values: List[np.float32]

    @dataobject
    NpFloat64Sequence(DataObjectBase):
        values: List[np.float64]

    data: Union[
        PyFloatSequence,
        NpFloat32Sequence,
        NpFloat64Sequence,
    ]
```

caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,188 @@
...
+        return {"data": self.data}
+
+    def fill_proto(self, proto):
+        data_proto = getattr(proto, "data")
+        data_proto.extend(self._convert_to_floatvalue(self.data))
+
+        return proto
+
+    @classmethod
+    def from_proto(cls, proto):
```
gabe-l-hart, line 78: Is this defined so that we can short-circuit the automatic stuff?

gabe-l-hart: Same question for all of these, actually

caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,188 @@
...
+        for datum in json_str["data"]:
+            data.append(datum)
+        return cls(data=data)
+
+
+@dataobject(package="caikit_data_model.caikit_nlp")
+class EmbeddingResult(DataObjectBase):
+    """Data representation for an embedding matrix holding 2D vectors"""
+
+    data: List[Vector1D]
```
gabe-l-hart, line 119: Hm, interesting. Even if we go with the above where all elements of the sequence need to be of a common type, you could have sequences here that are (a) different type and (b) different length. Could that cause any problems?

gabe-l-hart (commented):
Some thoughts about the way the dataobject fields are laid out

caikit_nlp/data_model/embedding_vectors.py
gkumbhat: Mismatch in length of individual list can certainly be the case with this approach. So essentially 2 vectors may not have equal length. But for now, I don't think we need to guarantee that at data model level. Like if the data is in python `List[List[]]`, which is a type we are supporting here, then that is anyways a possibility. 🤔

caikit_nlp/data_model/embedding_vectors.py
gkumbhat: Good point. I'll make the change

Issue #43, opened 2023-06-20
Add tasks and interfaces for sentence content detection modules


evaline-ju:
## Description

As part of #35 , I want to add Huggingface-aligned interfaces for the sentence content detection modules, so that I can have interfaces can potentially work for multiple transformers models and/or implementations.

## Discussion

Current thinking is:
- Alignment with text classification task for classification-related modules
- Potential alignment with token classification task for any detection-related modules

## Acceptance Criteria

- [ ] Tasks and interfaces for proposed sentence content detection modules
- [ ] Unit tests present

PR #46, opened 2023-06-22
:sparkles: Text and token classification tasks and data models

evaline-ju:
Basic interfaces meant to align with Huggingface [token classification](https://huggingface.co/docs/transformers/tasks/token_classification#inference) and [text classification](https://huggingface.co/docs/transformers/tasks/sequence_classification) tasks

Closes: #43 

This shows the intended usage of the data models and tasks with the context of the sentence content classification modules

Screen Shot 2023-06-22 at 9 25 15 AM


caikit_nlp/data_model/classification.py
```diff
@@ -0,0 +1,52 @@
...
+# Standard
+from typing import List
+
+# First Party
+from caikit.core import DataObjectBase
+import caikit
+
+
+@caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
+class Classification(DataObjectBase):
```
gkumbhat, line 26: Question: Is the reason for not using terms like: `SequenceClassification` or `TextClassification` that we may want to use same data model object for classification tasks in other domain like vision later on?

caikit_nlp/data_model/classification.py
```diff
@@ -0,0 +1,52 @@
...
+# https://huggingface.co/docs/transformers/tasks/token_classification#inference
+# The field `word` does not necessarily correspond to a single "word",
+# and `entity` may not always be applicable beyond "entity" in the NER
+# (named entity recognition) sense
+@caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
+class TokenClassification(DataObjectBase):
+    start: int
+    end: int
+    word: str  # could be thought of as text
+    entity: str  # could be thought of as label
```
gkumbhat, line 46: Is the plan to not expose `entity_group` and essentially use `entity` for both entity and entity_group ?

caikit_nlp/data_model/classification.py
```diff
@@ -0,0 +1,52 @@
...
+    results: List[Classification]
+
+
+# NOTE: This is meant to align with the HuggingFace token classification task:
+# https://huggingface.co/docs/transformers/tasks/token_classification#inference
+# The field `word` does not necessarily correspond to a single "word",
+# and `entity` may not always be applicable beyond "entity" in the NER
+# (named entity recognition) sense
+@caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
+class TokenClassification(DataObjectBase):
```
gkumbhat, line 42: nit: Should we call this `TokenClassificationObject` and above one as `ClassificationObject` ?

caikit_nlp/data_model/text.py
```diff
@@ -0,0 +1,27 @@
...
+
+# First Party
+from caikit.core import DataObjectBase
+import caikit
+
+
+@caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
+class Span(DataObjectBase):
+    start: int
+    end: int
```
gkumbhat, line 26: nit: can we add a comment if the `end` is inclusive  or would you say thats implementation detail ?

gkumbhat (commented):
left some small questions, otherwise looks good

caikit_nlp/data_model/classification.py
evaline-ju: Yes that is what I had envisioned here, that label and score are not really domain specific e.g. https://huggingface.co/docs/transformers/tasks/image_classification. We could consider if this is reason for "promotion" to `caikit/caikit` later if this is to be reused.

caikit_nlp/data_model/classification.py
evaline-ju: That is the thought for now. Technically keeping `entity` here does not completely disallow for a future `entity_group` usage if like huggingface, we provide the option/ability to aggregate or disambiguate between entities

caikit_nlp/data_model/text.py
evaline-ju: I think that's a bit of an implementation detail since even if we write a comment here, this does not enforce whether or not it's inclusive. Best practice may be for it to not be inclusive, but users of this data model are certainly free to do whatever they want.

caikit_nlp/data_model/classification.py
evaline-ju: My take is that it's fairly implied these are all "objects" and would prefer not to make names longer 😄

caikit_nlp/data_model/classification.py
gkumbhat: sounds good

caikit_nlp/data_model/classification.py
gkumbhat: My concern is, not having `entity_group` doesn't align the DM with HF fully, specially for entities case and leaves a possibility of future use of `entity` to define `entity_groups`.

evaline-ju: sure, I can add it in addition

gkumbhat (approved):

PR #50, opened 2023-06-23
Support tgis stop params

gkumbhat:
### Changes
- Add `max new tokens` and `min new tokens` param for text generation with TGIS backend


alex-jw-brooks (approved):
Looks good! We might want to expose the params in the non-TGIS based modules as well, but that can be a separate PR 😄

Issue #56, opened 2023-06-28
Add streaming support for sentence-level classification


evaline-ju:
## Is your feature request related to a problem? Please describe.

In addition to #35, we want to support streaming for sentence-level classification, so that users can stream in text and receive a stream of token classification outputs.

Architecture is expected to be compatible with full streaming support in `caikit` here: https://github.com/caikit/caikit/milestone/11

## Describe the solution you'd like

- Streaming task for token classification
- Streaming implementations for token classification modules (sequence classification-transformer-compatible sentence classification)
     - This will require sentence-detection ability for modules that require sentence splitting.
     - Spans will have to be tracked, for both modules that do sentence splitting and without

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #58, opened 2023-07-06
:sparkles: Add streaming support for filtered span classification

evaline-ju:
#57 without the one DCO miss :( 

Changes
- Generalize the "sequence classifier" for filtered span classification to work for classifiers that do either token classification or text classification
- Add new DM for streaming filtered span classification
- Add sentence splitting tokenization module using simple regex approach
- Implement bidirectional-streaming support for `FilteredSpanClassification` in `run_bidi_stream` function.

Closes: #56 

Future TODOs (to be addressed in separate PR):
- Bump caikit (likely minor) version for actual streaming support


gkumbhat (approved):
LGTM

PR #62, opened 2023-07-07
:bug: Fix issues with stream-span processing with trailing or leading whitespaces

gkumbhat:
### Changes
- Fix issue where if a sentence in the middle of the stream has trailing or leading whitespaces, then the offsets would be off and we may miss processing 1 character (start of the sentence for next sentence)


evaline-ju (approved):
LGTM - thanks!

Issue #71, opened 2023-07-11
Add bare-minimum implementation of fine-tuning using HF Trainer


gkumbhat:
## Description

As a `caikit-nlp` developer, I want to add a simple bare-minimum fine-tuning implementation that defines how we can use HF Trainer for fine-tuning.

This implementation would be phase1 implementation for fine-tuning that will allow us to iteratively improve, clean and add features to the fine-tuning module.

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #72, opened 2023-07-11
Add finetuning generation

gkumbhat:
closes: #71 

### Changes

- Add fine-tuning module that uses HF Trainer to showcase how we can use it to fine-tune a model.

### Notes:
- This is purposefully only using 1 type of trainer and supporting 1 type of models. The purpose is to lay out small changes and iteratively improve to avoid giant PRs.
- For this PR to work fully, we need to update `transformers` version to include the bugfix merged in the PR https://github.com/huggingface/transformers/pull/24759

### DIFF
--- a/caikit_nlp/modules/text_generation/fine_tuning.py
+++ b/caikit_nlp/modules/text_generation/fine_tuning.py
@@ -13,9 +13,6 @@
 # limitations under the License.
 
 
-# Standard
-import os
-
 # Third Party
 from torch.utils.data import IterableDataset
 from transformers import (
@@ -29,22 +26,14 @@ from transformers import (
 
 # First Party
 from caikit.core.data_model import DataStream
-from caikit.core.module_backends import BackendBase, backend_types
-from caikit.core.modules import ModuleBase, ModuleConfig, ModuleSaver, module
+from caikit.core.modules import ModuleBase, module
 from caikit.core.toolkit import error_handler, wip_decorator
-from caikit_tgis_backend import TGISBackend
-from caikit_tgis_backend.protobufs import generation_pb2
 import alog
 
 # Local
 from ...data_model import GeneratedResult, GenerationTrainRecord
-from ...resources.pretrained_model import (
-    HFAutoCausalLM,
-    HFAutoSeq2SeqLM,
-    PretrainedModelBase,
-)
 from ...toolkit.data_stream_wrapper import SimpleIterableStreamWrapper
-from ...toolkit.data_type_utils import get_torch_dtype, str_to_torch_dtype
+from ...toolkit.data_type_utils import get_torch_dtype
 from .text_generation_task import TextGenerationTask
 
 log = alog.use_channel("TXT_GEN")
@@ -68,10 +57,12 @@ class FineTuning(ModuleBase):
         super().__init__()
 
         self.tokenizer = tokenizer
-        # NOTE: self.model here can also be HF trainer. This is because if we have just trained the model
-        # then the models weights might be available in different devices (and configuration), depending on
-        # how it was trained. For now (July 10, 2023), we are not trying to extract the model out from trainer
-        # itself, since that would require us to essentially save it or reconstruct it to do normal inferring.
+        # NOTE: self.model here can also be HF trainer. This is because
+        # if we have just trained the model then the models weights might be
+        # available in different devices (and configuration), depending on
+        # how it was trained. For now (July 10, 2023), we are not trying to
+        # extract the model out from trainer itself, since that would require
+        # us to essentially save it or reconstruct it to do normal inferring.
         self.model = model
 
     @classmethod
@@ -180,6 +171,7 @@ class FineTuning(ModuleBase):
             model=trainer,
         )
 
+    # pylint: disable=unused-argument
     def run(
         self, text, preserve_input_text=False, max_new_tokens=20, min_new_tokens=0
     ) -> "GeneratedResult":
@@ -221,7 +213,7 @@ class FineTuning(ModuleBase):
             )
 
         else:
-            NotImplementedError(
+            raise NotImplementedError(
                 "model prediction on pre-finetuned model currently not supported"
             )
 
--- a/caikit_nlp/modules/text_generation/fine_tuning.py
+++ b/caikit_nlp/modules/text_generation/fine_tuning.py
@@ -17,7 +17,8 @@
 import os
 
 # Third Party
-from transformers import AutoConfig
+from torch.utils.data import IterableDataset
+from transformers import AutoConfig, AutoTokenizer, DataCollatorForSeq2Seq, Seq2SeqTrainingArguments, Seq2SeqTrainer
 
 # First Party
 from caikit.core.data_model import DataStream
@@ -35,6 +36,8 @@ from ...resources.pretrained_model import (
     HFAutoSeq2SeqLM,
     PretrainedModelBase,
 )
+from ...toolkit.data_stream_wrapper import SimpleIterableStreamWrapper
+from ...toolkit.data_type_utils import get_torch_dtype, str_to_torch_dtype
 from .text_generation_task import TextGenerationTask
 
 log = alog.use_channel("TXT_GEN")
@@ -43,7 +46,7 @@ error = error_handler.get(log)
 
 # pylint: disable=too-many-lines,too-many-instance-attributes
 @module(
-    id="f9181353-4ccf-4572-bd1e-f12bcda26792",
+    id="28a81449-32ce-4be3-b688-545bde68f738",
     name="Text Generation",
     version="0.1.0",
     task=TextGenerationTask,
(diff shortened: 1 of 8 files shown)


tests/modules/text_generation/test_fine_tuning.py
```diff
@@ -0,0 +1,50 @@
...
+            [
+                GenerationTrainRecord(
+                    input="@foo what a cute dog!", output="no complaint"
+                ),
+                GenerationTrainRecord(
+                    input="@bar this is the worst idea ever.", output="complaint"
+                ),
+            ]
+        ),
+        "torch_dtype": torch.bfloat16,
```
alex-jw-brooks, line 42: Small thing, but can we use float32 here, since bfloat16 is not supported on all devices? Unsupported type fallbacks should be handled in other tests, but it would be nice to make sure this test runs identically on most devices

tests/modules/text_generation/test_fine_tuning.py
```diff
@@ -0,0 +1,50 @@
...
+    }
+    model = FineTuning.train(
+        **train_kwargs
+    )
+    assert isinstance(model.model, Trainer)
+    # Ensure that we can get something out of it
+    pred = model.run("@bar what a cute cat!")
```
alex-jw-brooks, line 49: Can you add a check for the type of `model`?

tests/modules/text_generation/test_fine_tuning.py
```diff
@@ -0,0 +1,50 @@
...
+import torch
+from transformers import Trainer
+
+# First Party
+import caikit
+
+# Local
+from caikit_nlp.data_model import GeneratedResult, GenerationTrainRecord
+from caikit_nlp.modules.text_generation import FineTuning
+from caikit_nlp.resources.pretrained_model import HFAutoCausalLM, HFAutoSeq2SeqLM
```
alex-jw-brooks, line 12: Nit: some unused imports here, e.g., `HFAutoCausalLM`

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -0,0 +1,262 @@
...
+                max_new_tokens=max_new_tokens,
+                min_new_tokens=min_new_tokens,
+            )
+
+            generated_text = self.tokenizer.batch_decode(
+                generated_tokens.detach().cpu().numpy(), skip_special_tokens=True
+            )
+
+        else:
+            raise NotImplementedError(
```
alex-jw-brooks, line 220: Error types should be wrapped by the error handler!

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -0,0 +1,262 @@
...
+from caikit.core.toolkit import error_handler, wip_decorator
+import alog
+
+# Local
+from ...data_model import GeneratedResult, GenerationTrainRecord
+from ...toolkit.data_stream_wrapper import SimpleIterableStreamWrapper
+from ...toolkit.data_type_utils import get_torch_dtype
+from .text_generation_task import TextGenerationTask
+
+log = alog.use_channel("TXT_GEN")
```
alex-jw-brooks, line 39:
```suggestion
log = alog.use_channel("FIN_TUN_GEN")
```

So that we don't have collisions in logging channels with the separate text generation module

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -0,0 +1,262 @@
...
+        # available in different devices (and configuration), depending on
+        # how it was trained. For now (July 10, 2023), we are not trying to
+        # extract the model out from trainer itself, since that would require
+        # us to essentially save it or reconstruct it to do normal inferring.
+        self.model = model
+
+    @classmethod
+    def train(
+        cls,
+        base_model: str,  # TODO: Union[str, PretrainedModelBase]
```
alex-jw-brooks, line 71: Somewhat unrelated - I can't remember the exact reason unions were disallowed in our type hints for a little while, maybe something for caikit runtime type hint inspection and unions being unsupported at the time. I think that some work has been going on around that recently though - do you know if we can use them yet?

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -0,0 +1,262 @@
...
+    ):
+        """
+        # FIXME: Below is currently configured for Seq2Seq only
+        """
+
+        torch_dtype = get_torch_dtype(torch_dtype)
+
+        ## Load base model
+        if isinstance(base_model, str):
+            model_config = AutoConfig.from_pretrained(base_model)
```
alex-jw-brooks, line 91: Hmm - this is duplicated in a few places, e.g., [here](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/text_generation.py#L119). Can you add a note to consolidate these into a common base class in the future or something? Would like to try to keep tabs on some of these things for tech health

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -0,0 +1,262 @@
...
+        cls,
+        base_model: str,  # TODO: Union[str, PretrainedModelBase]
+        train_stream: DataStream[GenerationTrainRecord],
+        torch_dtype: str = None,  # TODO: Optional[Union[torch.dtype, str]]
+        max_source_length: int = 256,
+        max_target_length: int = 128,
+        batch_size: int = 8,
+        num_epochs: int = 5,
+        accumulate_steps: int = 32,
+        shuffle: bool = True,
```
alex-jw-brooks, line 79: Should `shuffle` really be parameterized here? Might be better to just always use True for this. E.g., this is what we do in the wrapper for Peft prompt tuning ([here](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L730))

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -0,0 +1,262 @@
...
+            max_target_length=max_target_length,
+            shuffle=shuffle,
+        )
+
+        ## TODO: Fetch trainer from resource
+
+        # TODO: Make this whole thing configurable by end-users, by optionally accepting `training_args`
+        # as argument to this train function.
+        # TODO: Remove all the default used below and make them all configurable
+        training_args = Seq2SeqTrainingArguments(
```
alex-jw-brooks, line 123: I think I might misunderstand this - is this module not providing fine tuning support to causal LM also? Also, can we fallback to defaults in the trainer or something instead of hard coding default overrides here?

alex-jw-brooks (commented):
Thanks Gaurav! Some initial thoughts

caikit_nlp/modules/text_generation/fine_tuning.py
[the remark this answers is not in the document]
gkumbhat: agreed, this is only temporarily a separate module. I was imaging it to go into `.train` for the `text_generation` module and a lot of the core functionality go in toolkit or something. Like was planning to add a `get_trainer` function in individual resources which would move a lot of the code to individual resources so we would be able to basically use same logic in different `task` specific modules

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: hmm. honestly not sure, I guess certainly something we could try, but there is a lot of things influx in data models, so we might want things to be more stable and then make these changes

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: hmm. good question, I was thinking might be just better to provide option, but may be doesn't need to be 🤔

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: good idea. Will do that

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: this will eventually support trainings for various types of models, for this iteration, I just started with seq2seq. In next iteration, I'll move this whole logic to  individual `resources`, so we just do `.get_trainer()` on it.

I left a comment related to that [here](https://github.com/caikit/caikit-nlp/pull/72/files#diff-5e944359dd88fad4652c05a882bc242a6b567ded247a67ac24c9fc71ef9c263bR118)

caikit_nlp/toolkit/data_stream_wrapper.py
```diff
@@ -48,10 +50,26 @@ def __init__(self, stream, shuffle, buffer_size=None):
...
+        # each worker
+        else:  # in a worker process
+            # split workload
+            per_worker = int(
+                math.ceil((self.end - self.start) / float(worker_info.num_workers))
+            )
+            worker_id = worker_info.id
+            iter_start = self.start + worker_id * per_worker
+            iter_end = min(iter_start + per_worker, self.end)
+        return iter(range(iter_start, iter_end))
```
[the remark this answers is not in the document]
gkumbhat: you mean add for multi-device case? that part is tricky since we won't have multi-device available in testing environment. I guess, I can patch the `get_worker_info` from transformers to write the test for this 🤔

tests/modules/text_generation/test_fine_tuning.py
gkumbhat: good point!

tests/modules/text_generation/test_fine_tuning.py
gkumbhat: you mean check for `model.model` ?

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: changed to  always be `True`

caikit_nlp/modules/text_generation/fine_tuning.py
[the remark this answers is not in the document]
alex-jw-brooks: Awesome, I think that sounds like a good plan

alex-jw-brooks (approved):
Approving for now - I think the longer term plan for this sounds good, and it's fine to keep up while it's in wip state. Thanks @gkumbhat!

Issue #66, opened 2023-07-07
Local inferencing with text generation module


Ssukriti:
## Description

As a caikit-nlp user, I want to be able to run text generation module locally on my machine, so that I can execute prompt engineering without depending on TGIS backend.

## Discussion

1. I will contribute block that will allow us to execute text generation module on local backend without TGIS . The block will use HF generate API.

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #79, opened 2023-07-13
text generation local block

Ssukriti:
Closes: https://github.com/caikit/caikit-nlp/issues/66

This PR adds support for local inferencing of text generation block. It will enable the block to be used in embed libraries


Ssukriti:
I need to have a short discussion with  @gkumbhat on these topics (I have set up a call):
1. Would it be a breaking change to rename existing TextGeneration class name to TextGenerationTGIS ?
2. What is role of `bos_token, sep_token, pad_token` ?
3. output with `preserve_input_text` flag set to True
4. refactoring of code to allow re-use of common functions

caikit_nlp/modules/text_generation/text_generation_local.py
```diff
@@ -0,0 +1,247 @@
...
+    ):
+        super().__init__()
+
+        error.type_check("<NLP48137045E>", str, allow_none=True, eos_token=eos_token)
+        self.base_model = base_model
+        self.base_model_name = base_model_name
+
+        # Set _model_loaded as False by default. This will only get set to True if
+        # we enable the tgis_backend and we are able to fetch the client successfully.
+        self._model_loaded = False
```
alex-jw-brooks, line 68: This attribute and comment are copy pasted from the TGIS remote module, right?

caikit_nlp/modules/text_generation/text_generation_local.py
```diff
@@ -0,0 +1,247 @@
...
+        Args:
+            model_path: str
+                Path to the model to be loaded.
+        Returns:
+            TextGeneration
+                Instance of this class built from the on disk model.
+        """
+
+        config = ModuleConfig.load(model_path)
+        base_model_path = config.get("artifact_path", "")
```
alex-jw-brooks, line 133: Why is there a fallback here? Do we expect a caikit config file to be added directly to HF models?

tests/modules/text_generation/test_text_generation_tgis.py
```diff
@@ -9,21 +9,36 @@
...
     StubTGISBackend,
     StubTGISClient,
 )
 
+### Stub Modules
+
+# Helper stubs / mocks; we use these to patch caikit so that we don't actually
+# test the TGIS backend directly, and instead stub the client and inspect the
+# args that we pass to it.
+class StubClient:
```
alex-jw-brooks, line 25: Looks like this is unused - I assume this is from StubTGISClient duplication before it was moved into the common fixture. Can we remove it?

alex-jw-brooks (approved):
A few small things, but this is great!

caikit_nlp/modules/text_generation/text_generation_local.py
gkumbhat: good point. There is a potential use-case of this in the realms of using HF models directly, but for the moment, that is not target, so it would be good to explicitly fail, instead of magically work, I'll remove this.

@Ssukriti please let me know if you had other thoughts?

tests/modules/text_generation/test_text_generation_tgis.py
gkumbhat: yep. seems like leftover from rebase. removing it

Issue #3, opened 2023-06-02
Remove device arg from .run()


alex-jw-brooks:
# Background
Currently, the Peft prompt tuning block takes a device override argument for inference ([here](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L157)). This is silly since it's not actually changing the `model` device. Changing the model device at inference time is also a bit unsafe from a parallelism perspective, so we should:

- Not allow the device to be specified by .run()
- Only let the device be specific at:
- [ ] Train time (done)
- [ ] Load time (needs to be added) 

Note: this should be tested with a Causal LM model, because Peft still doesn't support `.generate()` for seq2seq models.

### Acceptance Criteria
- [ ] Device _cannot_ be specified at inference time
- [x] Device can be specified at train time
- [ ] Device can be specified at load time
- [ ] Tests added

rawkintrevo:
@alex-jw-brooks during refactoring what ever "here" links to broke, could you please update at your convenience?

alex-jw-brooks:
Fixed - thanks!

chakrn:
Closing per @rawkintrevo

PR #80, opened 2023-07-13
Add support causalm finetune

### LINKED ISSUE #77: Add support for causal lm to fine-tuning
## Description

The 1st iteration of fine-tuning module was designed around seq2seq, in this second iteration, we wan to expand it to be able to work with causal-lm models.

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

gkumbhat:
Closes #77 

### Changes
- Add support for causal-lm fine-tuning
- Move trainer and training argument configuration to respective resource folder to allow easy selection and configuration in tuning module.
- Fix unit test issue where having cuda device would make accelerate put calculations on GPU while rest of the data is on CPU thus raising error. This was fixed by adding `set_cpu_device` fixture which changes the cuda environment variable and patches `is_available` function in `torch.cuda`


caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
```diff
@@ -52,6 +52,7 @@ def build_task_tokenize_function(
         max_source_length: int,
         max_target_length: int,
         verbalizer: str,
+        task_ids: Union[None, int] = None,
```
alex-jw-brooks, line 55: A couple of questions!

1. What is the motivation for doing this? I.e., is it to avoid the cost of setting task IDs for prompt tuning etc, or is it to not hardcode 0?
2. Should this be an `int`? I know we generally only use 0 for MPT with single target task, but If this is being made generic to not hardcore 0, an iterable is probably needed right?

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -144,47 +141,31 @@ def train(
...
-            disable_tqdm=True,
-            # NOTE: Following not possible without save and eval strategy
-            # load_best_model_at_end=True,
+            "do_eval": False,
+            # "evaluation_strategy ": "epoch",
+            "learning_rate": lr,
+            "weight_decay": 0.01,
+            "save_total_limit": 3,
+            "push_to_hub": False,
+            "no_cuda": False,  # Default
```
alex-jw-brooks, line 157: Why set this instead of just letting the default?

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -81,6 +75,7 @@ def train(
         lr: float = 2e-5,
         # Directory where model predictions and checkpoints will be written
         checkpoint_dir: str = "/tmp",
+        **training_arguments,
```
alex-jw-brooks, line 78: This is better! Can you link the trainer args in the docstring through?

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -81,6 +75,7 @@ def train(
...
     ):
         """
         # FIXME: Below is currently configured for Seq2Seq only
```
alex-jw-brooks, line 81: This should be removed, right?

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -247,13 +228,27 @@ def run(
...
+            else:
+                # NOTE: Currently the default trainer doesn't support easy way to run individual
+                # samples without converting them into Datasets etc. There is a
+                # predict_with_generate flag, but it doesn't do anything.
+                # Applicable for transformers==4.31.0
+                error(
+                    "<NLP39984681E>",
+                    NotImplementedError(
+                        f"Generation on {type(self.model)} not support \
+                      currently! Please try saving and running this model in TGIS."
```
alex-jw-brooks, line 246: oof. Does exporting via the trainer save API + reloading give you a transformer model back? I wonder if it would be better to have the first inference call export and reload with a warning until we find something better / implement a causal LM trainer doing something similar. Slow feels better than completely broken here IMO.

Or, is there any way we can cast to the seq2seq trainer and leverage the generate API for that? I guess that probably doesn't handle shifting etc the same way...

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -282,28 +278,14 @@ def _preprocess_function(
...
-
-            return model_inputs
-
-        return SimpleIterableStreamWrapper(
-            train_stream.map(_tokenization_func), shuffle=shuffle
+        (
+            tokenize_function,
+            requires_unwrapping,
+        ) = base_model.build_task_tokenize_function(
+            tokenizer, max_source_length, max_target_length, verbalizer=""
```
alex-jw-brooks, line 285: I think this is a bug - verbalizer template should not be empty here, right? I.e., i.e., we need to specify the actual attr/text to use, otherwise things won't work correctly, since verbalizer template rendering requires us to explicitly specify where the source text is rendered out of too

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -282,28 +278,14 @@ def _preprocess_function(
...
         )
+        mapped_stream = train_stream.map(tokenize_function)
```
alex-jw-brooks, line 287: Just a note for later, but this has the same performance pitfall we had previously discussed, which is also mentioned here https://github.com/caikit/caikit-nlp/pull/100#issuecomment-1651229995

If we revisit it in the PT module, it would be nice to handle it here also, since the solution will be the same

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -199,8 +196,8 @@ def run(
         verbalized_text = render_verbalizer(self.verbalizer, {"input": text})
         # Apply the tokenizer to the sample text & move to correct device
         tok_tensors = self.tokenizer(verbalized_text, return_tensors="pt")
-        device = PeftPromptTuning._get_device(device)
-        inputs = {k: v.to(device) for k, v in tok_tensors.items()}
+
+        inputs = {k: v.to(self.model.device) for k, v in tok_tensors.items()}
```
alex-jw-brooks, line 200: FYI @rawkintrevo is actually making this change in a separate PR (it's this issue https://github.com/caikit/caikit-nlp/issues/3). Can we put it back as part of this PR and use his when it's ready instead? Since this PR is primarily targeting fine tuning anyway

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -1066,7 +1068,13 @@ def _execute_train_loop(
             num_training_steps=(len(train_dataloader) * num_epochs),
         )
         # Configure accelerator for gradient accumulation
-        accelerator = Accelerator(gradient_accumulation_steps=accumulate_steps)
+        accelerator_args = {
+            "gradient_accumulation_steps": accumulate_steps,
+            "device_placement": True,
+        }
+
+        accelerator = Accelerator(**accelerator_args)
```
alex-jw-brooks, line 1076: why build a separate dict here?

tests/fixtures/__init__.py
```diff
@@ -32,6 +32,20 @@
 SEQ2SEQ_LM_MODEL = os.path.join(TINY_MODELS_DIR, "T5ForConditionalGeneration")
 
 
+@pytest.fixture()
+def set_cpu_device(request):
```
alex-jw-brooks, line 36: Nice - thanks for adding this

caikit_nlp/resources/pretrained_model/hf_auto_seq2seq_lm.py
```diff
@@ -68,12 +74,67 @@ def get_num_transformers_submodules(
...
+        eval_dataset: Union[IterableDataset, None] = None,
+        optimizers=(None, None),
+        **kwargs
+    ):
+        """
+        NOTE: following parameters are not supported currently:
+            1. model_init
+            2. compute_metrics
+            3. callbacks
+            4. preprocess_logits_for_metrics
```
alex-jw-brooks, line 89: Can you add arg descriptions here (at least for non kwargs)?

caikit_nlp/resources/pretrained_model/base.py
```diff
@@ -233,6 +239,57 @@ def save(
...
+        optimizers=(None, None),
+        **kwargs,
+    ):
+        """
+        NOTE: following parameters are not supported currently:
+            1. model_init
+            2. compute_metrics
+            3. callbacks
+            4. preprocess_logits_for_metrics
+        """
```
alex-jw-brooks, line 255: Same questions about documenting the kwargs here in the docstring (at least the nonexpanded ones). I assume the other one probably needs it also

alex-jw-brooks (changes requested):
Thanks @gkumbhat, looks good! Just a few things.

Also a bit of a side note, do you think we should remove support (for now) for `HFAutoSequenceClassifier `? Seems like it's effectively unusable between trainer changes & tokenizer builder stuff, it's kind of confusing to have it there when we don't enable it for anything

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: This was just to make some settings explicit, to avoid confusion later on with other parameters.

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: yeah, I think converting the seq2seq could land with weird mismatch issues. 

Saving and reloading is certainly an option. It would simplify this block of `run` function entirely. But could be more inefficient, since the model is already on appropriate devices at this point, so loading them again, we would loose the distribution, which is mainly what I was trying to persist here. 

But certainly, not having a solution of causal lm would not be great.

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: actually, idea here was to make verbalizer optional, since it won't be configured for fine-tuning currently.

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: true, this should easily be able to take improvements from there

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: ah true. I had to make this change to make some tests pass 😄  but yes, can change it back.

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: I was playing with some optional parameter regarding `cpu=True`.. But that didn't work well, so removed that.. So this is kinda left over from that.. Will switch it back to direct arguments instead of separate dict.

caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
gkumbhat: 1. `task_ids` doesn't work for non peft models actually and we get error from `transformers` when we try to do training. So this would become problematic for fine-tuning. So I made this field optional to re-use the same tokenization logic.
2. I thought of iterable, but then we would have to figure out how to iterate of tasks, since presumably either we would want to map the tasks_ids with training input  or duplicate bunch of task_ids with each of the training input. For current use-cases, we actually need this to be hardcoded, so thought of moving this to prompt tuning layer itself 🤔

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: yep. good catch! Will remove this

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -78,12 +87,49 @@ def train(
...
     ):
         """
-        # FIXME: Below is currently configured for Seq2Seq only
+        Fine-tune a CausalLM or Seq2seq text generation model.
+
+        Args:
+            base_model:  Union[str, caikit_nlp.resources.pretrained_model.base.PretrainedModelBase]
+                Base resource model used for underlying generation.
+            train_stream: DataStream[GenerationTrainRecord] or DataStream[ClassificationTrainRecord]
+                Data to be used for training the prompt vectors of the generation model.
```
alex-jw-brooks, line 103:
```suggestion
                Data to be used for fine-tuning the generation model.
```

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -144,47 +197,33 @@ def train(
...
+            "push_to_hub": False,
+            "no_cuda": False,  # Default
+            "remove_unused_columns": False,
+            "dataloader_pin_memory": False,
+            "gradient_accumulation_steps": accumulate_steps,
+            "eval_accumulation_steps": accumulate_steps,
             # eval_steps=1,
+            # load_best_model_at_end
+            **training_arguments,
             **dtype_based_params,
```
alex-jw-brooks, line 222: Might be a nice good first issue in the future to cleanly make sure there aren't collisions in these expanded dicts, but for now we can leave it

caikit_nlp/resources/pretrained_model/base.py
```diff
@@ -233,6 +239,61 @@ def save(
...
+    def get_trainer(
+        self,
+        train_dataset: IterableDataset,
+        eval_dataset: Union[IterableDataset, None] = None,
+        optimizers=(None, None),
+        **kwargs,
+    ):
+        """
+        Args:
+            *kwargs: arguments supported by HF TrainingArguments:
```
alex-jw-brooks, line 251:
```suggestion
            **kwargs: arguments supported by HF TrainingArguments:
```

caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
alex-jw-brooks: Cool sounds great. Thanks!

alex-jw-brooks (approved):
Looks awesome! Some small typos and stuff, but LGTM

caikit_nlp/modules/text_generation/fine_tuning.py
gkumbhat: good idea

Issue #75, opened 2023-07-12
Evaluate fine-tuning module on summarization task


gkumbhat:
## Description

As a caikit-nlp developer, I want to make sure fine-tuning on certain module provides "reasonable" quality model to make sure its working as expected

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #84, opened 2023-07-18
Add eval fine tuning

gkumbhat:
Closes: #75 

### Changes
- Update transformers to latest version that enables bugfix for Trainer prediction
- Add [samsum](https://huggingface.co/datasets/samsum) and [billsum](https://huggingface.co/datasets/billsum) dataset from huggingface to test out summarization
- Change default input text length to support summarization
- Fix prediction generation batch decoding to single output mapping issue
- Add evaluation in fine-tuning notebook

### Test
1. Unit test: `tox -e py39 -- tests`
2. Evaluation example:
    ```python
   ALLOW_DOWNLOADS=true python3 run_fine_tuning.py --dataset "billsum"  --model_name t5-base --num_epochs 4 --output_dir tmp/with_loss_log --batch_size=8 --accumulate_steps 16 --max_source_length 1024 --metric rouge --torch_dtype float16
    ```

### DIFF
--- a/caikit_nlp/modules/text_generation/fine_tuning.py
+++ b/caikit_nlp/modules/text_generation/fine_tuning.py
@@ -162,8 +162,15 @@ class FineTuning(ModuleBase):
             dataloader_pin_memory=False,
             gradient_accumulation_steps=accumulate_steps,
             eval_accumulation_steps=accumulate_steps,
+            logging_strategy="epoch",
+            disable_tqdm=True,
+            # NOTE: Following not possible without save and eval strategy
+            # load_best_model_at_end=True,
             # eval_steps=1,
             **dtype_based_params,
+            ## TODO: Make below configurable
+            # fsdp="full_shard auto_wrap",
+            # local_rank=0,
         )
 
         data_collator = DataCollatorForSeq2Seq(
--- a/caikit_nlp/modules/text_generation/fine_tuning.py
+++ b/caikit_nlp/modules/text_generation/fine_tuning.py
@@ -23,6 +23,7 @@ from transformers import (
     Seq2SeqTrainingArguments,
     Trainer,
 )
+import torch
 
 # First Party
 from caikit.core.data_model import DataStream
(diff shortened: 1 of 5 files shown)


examples/utils.py
```diff
@@ -80,9 +80,10 @@ def get_distributed_model(model_path):
         kill_tgis_container_if_exists()
 
     # TODO: Enforce validation that TGIS is mounting the same model type
-    _CONFIGURED_BACKENDS.clear()
     caikit.configure(
-        config_dict={"module_backends": {"priority": [TGISBackend.backend_type]}}
+        config_dict={
+            "module_backends": {"load_priority": [{"type": TGISBackend.backend_type}]}
```
evaline-ju, line 85: I wanted to try to run this but noticed these config changes and the `caikit.core.module_backends.module_backend_config` import appear to work with a `caikit` version prior to `0.10.0` (the current lower bound) - what version was this tested with? The caikit module backends appeared reorganized for `0.10.0` - https://github.com/caikit/caikit/pull/275

examples/utils.py
```diff
@@ -226,6 +227,46 @@ def to_generation_fmt(x):
...
+    dataset = datasets.load_dataset("billsum", split="ca_test")
+    train_test_dataset = dataset.train_test_split(test_size=0.2)
+    # # Split the 10% test + valid into half test, half valid
+    # test_valid = train_test_valid_dataset['test'].train_test_split(test=0.5)
+
+    build_stream = lambda split: caikit.core.data_model.DataStream.from_iterable(
+        [to_generation_fmt(datum) for datum in train_test_dataset[split]]
+    )
+    train_stream = build_stream("train")
+    validation_stream = build_stream("test")
```
evaline-ju, line 245: wanted to confirm if the "test" here is intentional vs. "validation"? same for the above financial phrasebook and not accidental copypaste

caikit_nlp/modules/text_generation/fine_tuning.py
```diff
@@ -119,6 +120,23 @@ def train(
             shuffle=True,
         )
 
+        ### Dtype based processing
+        # NOTE: Following is not exhaustive list of all parameters
+        # for all dtypes
+        if torch_dtype == torch.float16:
```
evaline-ju, line 126: would it be possible to add small unit test(s) for these code paths or is the overhead for train/run pretty high right now even for unit tests?

gkumbhat: actually anything half precision is only available on GPUs, so the current unit tests setup won't work for this

examples/utils.py
gkumbhat: this was tested on `0.9.1`. Need to get this updated.

examples/utils.py
gkumbhat: Yep, since this dataset doesn't have "validation" set

evaline-ju (approved):
LGTM - anticipating the examples scripts will be fixed in https://github.com/caikit/caikit-nlp/pull/90

evaline-ju (approved):

PR #90, opened 2023-07-20
:wrench: Fix utils script to configure backends in the new way in updated caikit

gkumbhat:
### Changes
- fix backends configuration for example scripts to supported caikit updated API

### DIFF
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -23,7 +23,6 @@ import torch
 import transformers
 
 # First Party
-from caikit.core.module_backend_config import _CONFIGURED_BACKENDS, configure
 from caikit_tgis_backend import TGISBackend
 import alog
 import caikit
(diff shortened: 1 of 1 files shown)


examples/utils.py
```diff
@@ -80,11 +79,20 @@ def get_distributed_model(model_path):
         kill_tgis_container_if_exists()
 
     # TODO: Enforce validation that TGIS is mounting the same model type
-    _CONFIGURED_BACKENDS.clear()
-    caikit.configure(
-        config_dict={"module_backends": {"priority": [TGISBackend.backend_type]}}
+
+    caikit.config.configure(
```
evaline-ju, line 83: I thought just `caikit.configure(...` should still also work but no big deal

evaline-ju (approved):

examples/utils.py
gkumbhat: True. I didn't realize that, but good to know. Thanks

PR #91, opened 2023-07-20
:recycle: Use generated text result from caikit.interfaces

evaline-ju:
Reference: https://github.com/caikit/caikit/blob/main/caikit/interfaces/nlp/data_model/text_generation.py

### DIFF
--- a/caikit_nlp/config/config.yml
+++ b/caikit_nlp/config/config.yml
@@ -23,7 +23,3 @@ source_prompt_base: ""
 
 runtime:
   library: caikit_nlp
-  service_generation:
-    primitive_data_model_types:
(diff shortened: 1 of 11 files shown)


caikit_nlp/config/config.yml
```diff
@@ -23,7 +23,3 @@ source_prompt_base: ""
 
 runtime:
   library: caikit_nlp
-  service_generation:
```
evaline-ju, line 26: This config was removed in an earlier caikit version: https://github.com/caikit/caikit/releases/tag/v0.7.0

gkumbhat (approved):
Thanks Evaline. LGTM

PR #93, opened 2023-07-21
Api dumps

gabe-l-hart:
## Description

This PR adds a script to dump the HTTP and gRPC APIs. It also bumps the `caikit` bounds to support HTTP


gkumbhat (approved):
Thanks Gabe 🥳

Issue #70, opened 2023-07-10
Simplify Prompt Tuning Tokenization


alex-jw-brooks:
## Is your feature request related to a problem? Please describe.
Currently, the way we do tokenization in the prompt tuning module has some problems. We effectively build a different tokenization function based on the type of auto model we are creating ([here](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L880)), but this is almost certainly way more complicated than it needs to be.

As part of the effort to add multiGPU support for prompt tuned models, we should simplify the tokenization logic and consider moving to HF trainers.

## Describe the solution you'd like
- Simple tokenization logic with no manual padding logic added; tokenizers/transformers should already handle this
- (Ideally) No custom logic for different tokenizing/data prep on different automodels outside of capabilities in the library where needed, e.g., different data collators 
- A short example illustrating that the tokenized output of the simplified result approximately matches the existing code or an explanation of why things have changed, with some reasonable justification for why it won't lower quality of the prompt tuned models

PR #95, opened 2023-07-21
Lang modeling

alex-jw-brooks:
This PR does a few things:
- Rewrites tokenization logic to more accurately reflect the objective for causal language modeling
- Utilizes the language modeling data collator for Causal LMs
- Pulls tokenization logic out into the toolkit & adds a builder to create the tokenization funcs, which are now leveraged in the Peft prompt tuning block
- Adds explicit tests for builder logic, validating the attention masks, etc

In terms of the language modeling objective - the way we handle this for source / target strings is to effectively generate a language modeling type response for every token in the target string.

Say we have the following:
```
source: "I am a" -> ["I", "am", "a"]
target: "target string" -> ["target", "string"]
```

When we apply the builder and map the tokenization func, we get a stream of streams, which can be flattened into the following:
```
["I", "am", "a"] -> "target" # Mask: [1, 1, 1, 1, 0]; don't attend to 'string'; we don't care about it yet
["I", "am", "a", "target"] -> "string" # Mask: [1, 1, 1, 1, 1]l here we attend to 'string' and treat 'target' as part of our source
```

Note that we also use dynamic padding via the collator to optimize dataset processing, which is more efficient, and is more in line with collator patterns. We also don't do anything specific for left/right padding anymore, which is great, since there were some issues with padding side in the previous tokenization func implementation.

It also fixes an issue in the truncation logic for Causal LM, since we were previously truncating our target sequence IDs with the source sequence limit ([here](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L1056C24-L1056C46)).

closes:  https://github.com/caikit/caikit-nlp/issues/70.


caikit_nlp/toolkit/data_prep_utils.py
```diff
@@ -0,0 +1,163 @@
...
+            tokenization / preprocessing behavior.
+
+    Returns:
+        Tuple(Callable, bool)
+            Mappable tokenize function to be applied to a training stream and bool indicating
+            whether or not the stream needs to be unwrapped, i.e., each sample yields a stream
+            of 1+ samples.
+    """
+
+    def tokenize_function_language_model(
```
gkumbhat, line 68: Should we move the tokenization function to respective resources?

gkumbhat: ah yes, the static method route sound good!

alex-jw-brooks: Cool sounds good - it's not beautiful, but I did refactor it + the tests into the resource classes 😄 may be a good idea to clean this up a bit over time, but at least we can preserve the resource APIs throughout future refactors on it

gkumbhat (approved):
LGTM

PR #99, opened 2023-07-25
:bug: Fix autoconfig always fetching info from model-hub

gkumbhat:
### Changes
- Currently `Autoconfig` is by default going off of HF hub and fetching model configurations, which may not always be required depending on the usecase and may create confusion if the model version changed on HF Hub. This PR makes this configurable with the `ALLOW_DOWNLOADS` flag


alex-jw-brooks (approved):

alex-jw-brooks:
Thanks, gaurav!

PR #107, opened 2023-07-27
Remote tgis text gen

gabe-l-hart:
## Description

This PR makes updates to `peft_tgis_remote` and `text_generation`, the two `text_generation` module implementations that use a `TGISBackend`, to make them work with remote TGIS connections and no local artifacts. The changes include:

* Updating the `save/load` logic in `TextGeneration` to support saving and loading without artifacts
* Update the `save/load` logic in `TextGeneration` to preserve the base model name
* Update `PeftPromptTuningTGIS` to call the backend's `load_prompt_artifacts` to get the artifacts "loaded" to TGIS. In practice, this just means copying them to a preconfigured "prompt_dir" that the `TGISBackend` is assuming is a shared volume mount with TGIS for the prompt cache.

## Testing

Aside from unit tests, I've tested this out with a locally running server and port-forwards to TGIS. The testing setup looked like:

1. Get the server up and running with port forards

    ```sh
    # Port forward the two models, one for a prompt and one for a "base" model
    k port-forward svc/llama2-70b-inference-server 8034:8033 &
    k port-forward svc/flan-t5-3b-inference-server 8033:8033 &

    # Run the server locally
    ./scripts/run_local.sh
    ```

2.  Get the server to load some models by copying the attached stub models into `models`

    [sample_models.tgz](https://github.com/caikit/caikit-nlp/files/12188929/sample_models.tgz)

4. Call the "base" model and see some results!

    ```sh
    curl -X 'POST' \
      'http://localhost:8080/api/v1/llama-70b/task/text-generation' \
      -H 'accept: application/json' \
      -H 'Content-Type: application/json' \
      -d '{
      "inputs": "this is a test"
    }'
    ```

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_tgis_remote.py
+++ b/caikit_nlp/modules/text_generation/peft_tgis_remote.py
@@ -15,7 +15,7 @@
 prompt vectors in TGIS generation requests.
 """
 # Standard
-from typing import Iterable
+from typing import Iterable, List, Optional
 import os
 
 # First Party
@@ -46,12 +46,13 @@ class PeftPromptTuningTGIS(ModuleBase):
 
     def __init__(
         self,
-        base_model_name,
-        prompt_cache_id,
-        eos_token,
-        verbalizer,
-        enable_backend=True,
-        tgis_backend=None,
+        base_model_name: str,
+        prompt_cache_id: str,
+        eos_token: str,
+        verbalizer: str,
+        enable_backend: bool = True,
+        tgis_backend: Optional[TGISBackend] = None,
+        prompt_artifacts: Optional[List[str]] = None,
     ) -> None:
         super().__init__()
         # Configure the internal client
@@ -62,6 +63,12 @@ class PeftPromptTuningTGIS(ModuleBase):
             # get_client will also launch a local TGIS process and get the model
             # loaded when using the local TGIS backend
             self._client = tgis_backend.get_client(base_model_name)
+
+            # Tell the backend to load all of the available prompt files
+            if prompt_artifacts:
+                tgis_backend.load_prompt_artifacts(
+                    base_model_name, prompt_cache_id, *prompt_artifacts
+                )
 
         self.base_model_name = base_model_name
         self._prompt_cache_id = prompt_cache_id
@@ -97,7 +104,7 @@ class PeftPromptTuningTGIS(ModuleBase):
         config = ModuleConfig.load(model_path)
         eos_token = config.eos_token
         verbalizer = config.verbalizer
-        dir_name = os.path.split(model_path)[-1]
+        dir_name = os.path.basename(model_path)
         # NOTE: base_model_name is used as "model_id" when calling to TGIS backend
         base_model_name = config.get("base_model_name", "")
         prompt_cache_id = dir_name
@@ -111,12 +118,24 @@ class PeftPromptTuningTGIS(ModuleBase):
         # we convert make it valid json compatible dict (aka doesn't have non string keys)
         log.debug("Prompt ID: %s", prompt_cache_id)
         log.debug("TGIS model ID: %s", base_model_name)
+
+        # Get all the valid prompt artifact files so they can be loaded after
+        # the connection is established
+        prompt_artifacts = [
+            os.path.join(model_path, config.get(config_key))
+            for config_key in [
+                PeftPromptTuning._ENCODER_KEY.name,
+                PeftPromptTuning._DECODER_KEY.name,
+            ]
+            if config.get(config_key)
+        ]
         return cls(
             base_model_name,
             prompt_cache_id,
             eos_token,
             verbalizer,
             tgis_backend=load_backend,
+            prompt_artifacts=prompt_artifacts,
         )
 
     def save(self, model_path: str):
--- a/caikit_nlp/modules/text_generation/text_generation.py
+++ b/caikit_nlp/modules/text_generation/text_generation.py
@@ -161,15 +161,11 @@ class TextGeneration(ModuleBase):
             model_path: str
                 Folder to save text-generation caikit model
         """
-        saver = ModuleSaver(
-            self,
-            model_path=model_path,
-        )
+        saver = ModuleSaver(self, model_path=model_path)
         with saver:
-            artifacts_dir = "artifacts"
             saver.update_config(
                 {
-                    "artifact_path": artifacts_dir,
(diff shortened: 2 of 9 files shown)


caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -153,22 +153,21 @@ def save(self, model_path):
...
                     "bos_token": self._bos_token,
                     "sep_token": self._sep_token,
                     "eos_token": self._eos_token,
                     "pad_token": self._pad_token,
                 }
             )
             if self.base_model:
+                artifacts_dir = "artifacts"
+                log.debug("Saving model artifacts to %s", artifacts_dir)
+                saver.update_config({"artifact_path": artifacts_dir})
```
alex-jw-brooks, line 170: I know this is kind of residual from what was previously there, but I wonder if this should be renamed, since it's pretty nondescriptive

scripts/run_local.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+mkdir -p models
```
alex-jw-brooks, line 2: Should this be created after the `cd` on line 4 so that we don't have random model dirs floating around if we run it in different dirs?

caikit_nlp/modules/text_generation/text_generation.py
gabe-l-hart: I think this would be a breaking change with existing models. There aren't any real ones that I know of at this point, but I also don't think `artifact_path` is too bad (it's the path to the model artifacts!), so I'm inclined to leave it.

scripts/run_local.sh
gabe-l-hart: Good catch!

caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -63,6 +64,12 @@ def __init__(
             # loaded when using the local TGIS backend
             self._client = tgis_backend.get_client(base_model_name)
 
+            # Tell the backend to load all of the available prompt files
+            if prompt_artifacts:
+                tgis_backend.load_prompt_artifacts(
+                    base_model_name, prompt_cache_id, *prompt_artifacts
```
gkumbhat, line 70: am wondering if we need to specifically get the `prompt_artifacts` to the `load_prompt_artifacts` function, since the `load` function is going to be just copying over the file,  so instead of copying specific list of files, i.e `encoder.pt` and `decoder.pt` it might be better to copy the whole folder which will include the `config.yml` ? That way the shape of the "model" if looked inside TGIS container look similar to shape available on some storage system.

On the other side, if there are any artifacts, like "base_model" also stored in the "model_folder" passed on here, then the "load" can incur a lot of cost 🤔 

I guess, either way, changing this won't be API breaking, so may be we don't need to make decision on that, but may be for future 💭

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -161,22 +168,21 @@ def save(self, model_path):
             model_path: str
                 Folder to save text-generation caikit model
         """
-        saver = ModuleSaver(
-            self,
-            model_path=model_path,
-        )
+        saver = ModuleSaver(self, model_path=model_path)
```
gkumbhat, line 171: Actually not related to this PR, but `AutoConfig` at line 132 will always try to connect to the internet so can cause issues.  This was fixed in prompt tuning training, but not here. Might be worth including in this PR since it can potentially break model running. 

Solution: https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L393

caikit_nlp/modules/text_generation/text_generation.py
```diff
@@ -89,6 +89,14 @@ def __init__(
             self._client = tgis_backend.get_client(base_model_name)
             # mark that the model is loaded so that we can unload it later
             self._model_loaded = True
+            # Make sure that we either have a base model or TGIS is running as a
+            # remote-proxy
+            error.value_check(
+                "<NLP51672289E>",
+                self.base_model or not tgis_backend.local_tgis,
+                "Cannot run model {} with TGIS locally since it has no base artifacts",
```
gkumbhat, line 97: nice! 🥳

caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -111,12 +118,24 @@ def load(cls, model_path: str, load_backend: BackendBase) -> "PeftPromptTuningTG
...
+
+        # Get all the valid prompt artifact files so they can be loaded after
+        # the connection is established
+        prompt_artifacts = [
+            os.path.join(model_path, config.get(config_key))
+            for config_key in [
+                PeftPromptTuning._ENCODER_KEY.name,
+                PeftPromptTuning._DECODER_KEY.name,
+            ]
+            if config.get(config_key)
```
evaline-ju, line 130: q - I'm a bit confused how any these keys (ENCODER/DECODER) would be expected to end up in the module config without changes to `save`?

gkumbhat (commented):
Looks great. Thanks Gabe! One tiny suggestion unrelated to this work, but other than that some thoughts and nits

caikit_nlp/modules/text_generation/peft_tgis_remote.py
evaline-ju: just kidding I see the `base_module` save now

caikit_nlp/modules/text_generation/peft_tgis_remote.py
gabe-l-hart: Yeah, I thought about this, but figured that this is "prompt specific" and in the future if TGIS exposes CRUD on prompts, we'll only be sending the prompt data and not the whole model (TGIS should not know or care about `caikit` formatted models). The point about models that contain the base model artifacts is valid too. If we did want to do this, I'd say we would want to nest `encoder.pt`/`decoder.pt` in a subdirectory like `prompt_artifacts`, but then the flat model directory would not be able to function as-is in TGIS since it would have an extra layer of name nesting. Given all this, I'm inclined to leave it as is.

caikit_nlp/modules/text_generation/text_generation.py
gabe-l-hart: Nice, good catch!

gkumbhat (approved):
Looks great. Thanks Gabe!

PR #112, opened 2023-08-01
[CAIKIT-NLP-3] Remove device arg from .run

rawkintrevo:
Disabled calling of device functionality- but left argument in place (defaulted to none) to maintain backwards compatibility. Functionality no longer exists- if user sets device parameter a warning message is shown, otherwise device is detected manually.


rawkintrevo:
Can someone tell me what this lint thing is doing besides failling and making it impossible to update my code w/o a force push?

alex-jw-brooks:
Hey @rawkintrevo! There's instructions on how to set up & run the code formatter and linter [here](https://github.com/caikit/caikit-nlp/blob/main/CONTRIBUTING.md#coding-style). It looks like this is failing due to the black code formatter - running the format command and checking in the result should resolve it

rawkintrevo:
bump @gkumbhat / @alex-jw-brooks could i get a review and this merged?

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -195,6 +195,11 @@ def run(
             GeneratedTextResult
                 Generated text result produced by PEFT / Transformers.
         """
+        if device is not None:
+            log.warning(
+                "Specifying device is deprecated and ignored, please update your calling argument"
+            )
+        device = self._DETECT_DEVICE
```
gkumbhat, line 202: I think instead of this, we should fetch the device information from the `model` itself, otherwise there is still possibility of discrepancy. Can we instead do `self.model.device` to fetch the information which device the model is ?

gkumbhat (changes requested):
Thanks @rawkintrevo . Left some suggestions and also it seems this branch needs a rebase since its showing changes from previous PR as net new changes

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
rawkintrevo: The old line 177 set device this way, I assumed there was a reason for it?

rawkintrevo:
@gkumbhat updated the branch- I think it fell out of sync because it has been waiting for a review for ~2 weeks. Responded to your other comment.

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: Hmm. in grand scheme of things, it was sort of a bug 😄

alex-jw-brooks: Yeah, it actually is a bug since the device has to match the model, otherwise it'll throw because the tensors are on wrong devices. Probably the right behavior here is to remove this parameter from `.run` (since we are in `0.x` still) rather than keep it here, since the only way this parameter can currently be used should be just putting it on the model device

gkumbhat (approved):
This will need to be modified with the refactors coming in for #140 . Merging it in for now.

Issue #81, opened 2023-07-14
Consolidate text-generation module with fine-tuning module


gkumbhat:
## Description

In the first iteration we started with text_generation and fine_tuning being separate module, which becomes a bit confusing because of inoperability of the modules which are meant to perform same task. Additionally we would also need to add TGIS support for fine-tuning, further making these 2 modules similar. 

In this story, we want to combine these 2 modules to provide single interface

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

evaline-ju:
#96 did not cover fine tuning and #97 was closed in favor of this, so this issue will presumably cover consolidating output streaming functionality as well

PR #116, opened 2023-08-03
Consolidate tgis finetuning

gkumbhat:
closes: #81 

### Changes
- Remove fine-tuning module
- Move `.train` function and related pre-processing function from fine-tuning modules to text-generation local module
- Refactor save and load to provide portability between these
- Add tests


examples/utils.py
```diff
@@ -99,13 +101,17 @@ def get_distributed_model(model_path):
     # NOTE: bloom-560m is the default here because that's the default model used in our
     # text generation server hack script.
     model_name_override = os.getenv("MODEL_NAME", "bloom-560m")
```
alex-jw-brooks, line 103:
```suggestion
    model_name_override = os.getenv("MODEL_NAME", "bigscience/bloom-560m")
```
Just noticed that this is defaulting to the wrong name

examples/utils.py
```diff
@@ -99,13 +101,17 @@ def get_distributed_model(model_path):
...
-    loaded_base_model = dist_model.base_model_name
+    if hasattr(dist_model, "base_model_name"):
+        loaded_base_model = dist_model.base_model_name
+    else:
+        loaded_base_model = dist_model.model_name
     if not model_name_override.endswith(loaded_base_model):
-        raise ValueError(
+        log.error(
```
alex-jw-brooks, line 109: Why is this being made non fatal?

examples/run_fine_tuning.py
```diff
@@ -325,13 +334,34 @@ def export_model_preds(preds_file, predictions, validation_stream):
...
+        # Saving model
+        model.save(args.output_dir)
+
+        # Load model in TGIS
+        # HACK: export args.output_dir as MODEL_NAME for TGIS
+        # container to pick up automatically
+        os.environ["MODEL_DIR"] = os.path.dirname(args.output_dir)
+        os.environ[
+            "MODEL_NAME"
+        ] = f"/models/{os.path.basename(args.output_dir)}/artifacts"
```
alex-jw-brooks, line 348: Would be nice to use `os.path.join` here instead of hardcoding `/` for platform compatibility

examples/run_fine_tuning.py
```diff
@@ -325,13 +334,34 @@ def export_model_preds(preds_file, predictions, validation_stream):
...
 
+    if args.tgis:
+
+        # Saving model
+        model.save(args.output_dir)
+
+        # Load model in TGIS
+        # HACK: export args.output_dir as MODEL_NAME for TGIS
+        # container to pick up automatically
+        os.environ["MODEL_DIR"] = os.path.dirname(args.output_dir)
```
alex-jw-brooks, line 345: I might be missing it, but what is `MODEL_DIR` used for? This isn't a TGIS env var, right?

examples/run_fine_tuning.py
```diff
@@ -247,7 +253,12 @@ def get_model_preds_and_references(model, validation_stream):
     for datum in tqdm(validation_stream):
         # Local .run() currently prepends the input text to the generated string;
         # Ensure that we're just splitting the first predicted token & beyond.
-        raw_model_text = model.run(datum.input).text
+        # TEMP HACK to avoid too big input to TGIS problem:
+        try:
+            raw_model_text = model.run(datum.input).generated_text
+        except Exception as ex:
+            print(ex)
```
alex-jw-brooks, line 260: oof. Any way to make this more narrow, or is it like a general exception? Or, should we at least catch separately for in vs not in TGIS?

alex-jw-brooks (changes requested):
Thanks Gaurav, great work! Some thoughts, but in general this looks a lot better

alex-jw-brooks (approved):
LGTM!

tests/modules/text_generation/test_text_generation_tgis.py
```diff
@@ -79,6 +88,35 @@ def test_save_model_can_run():
...
+        "train_stream": caikit.core.data_model.DataStream.from_iterable(
+            [
+                GenerationTrainRecord(
+                    input="@foo what a cute dog!", output="no complaint"
+                )
+            ]
+        ),
+        "torch_dtype": torch.float32,
+    }
+    model = TextGeneration.train(**train_kwargs)
```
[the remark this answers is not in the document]
alex-jw-brooks: Yeah, it isn't necessary, but I do worry a bit that this will feel kind of strange coming from other caikit implementations, where it's very weird for train on one type to produce a loadable module of a different type. Let's punt that discussion for now though!

Issue #121, opened 2023-08-09
Test and fix service generation for fine-tuning / text-gen train module


gkumbhat:
## Description

Test out caikit service generation for fine-tuning module, i.e `text_generation_local` and fix any issues in successful service generation encountered in the process.

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #122, opened 2023-08-09
:bug: Fix training arguments for service generation to work correctly

gkumbhat:
closes: #121 

### Changes
- Fix training function argument to fix warnings and errors in caikit service generation.


joerunde (approved):
🌶️

PR #135, opened 2023-08-21
Closes #134

### LINKED ISSUE #134: Torch dtype not getting passsed- results in model being loaded twice if not float32
## Describe the bug

When finetuning, torch_dtype isn't passed to trainer, only float32 is used. If any type other than float32 is used, this results in the model being loaded on GPU twice, and for any but the smallest models, results in OOM errors. 

See this thread for more details. 
https://discuss.huggingface.co/t/why-is-uploaded-model-twice-the-size-of-actual-model/18782/4

## Platform

Please provide details about the environment you are using, including the following:

All

## Sample Code

Please include a minimal sample of the code that will (if possible) reproduce the bug in isolation

```bash
!python caikit-nlp/examples/run_fine_tuning.py --dataset "billsum"  --model_name tiiuae/falcon-rw-1b --num_epochs 1 --output_dir tmp/falcon-1b --batch_size=4 --accumulate_steps 8 --max_source_length 16 --max_target_length 16 --metric rouge --torch_dtype bfloat16
```

## Expected behavior

The user expects to be able to fit larger models in memory when using half precision (compared to full precision).

## Observed behavior

OOM Errors. It is possible to fit larger models in memory using float32 than it is using float16 or bfloat16

## Additional context

I lost my notes, but I'm pretty sure setting this line manually to `bfloat16` when calling bloat16 from the command line, made everything work as expected.

https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/resources/pretrained_model/base.py#L185

olson-ibm:
Simple bugfix to pass the Caikit NLP CLI `--dtype` argument to bootstrapping, allowing half precision models to be loaded `.from_pretrained()`

### DIFF
--- a/examples/run_fine_tuning.py
+++ b/examples/run_fine_tuning.py
@@ -313,7 +313,11 @@ if __name__ == "__main__":
 
     # Init the resource & Build the tuning config from our dataset/arg info
     print_colored("[Loading the base model resource...]")
-    base_model = model_type.bootstrap(args.model_name, tokenizer_name=args.model_name, torch_dtype=args.torch_dtype)
+    base_model = model_type.bootstrap(
+        args.model_name, 
+        tokenizer_name=args.model_name, 
+        torch_dtype=args.torch_dtype
+    )
 
     # Then actually train the model & save it
     print_colored("[Starting the training...]")
--- a/examples/run_fine_tuning.py
+++ b/examples/run_fine_tuning.py
@@ -313,7 +313,7 @@ if __name__ == "__main__":
 
     # Init the resource & Build the tuning config from our dataset/arg info
     print_colored("[Loading the base model resource...]")
-    base_model = model_type.bootstrap(args.model_name, tokenizer_name=args.model_name)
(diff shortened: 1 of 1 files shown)


gkumbhat (approved):
Thanks @olson-ibm for the fix. Looks good to me.

examples/run_fine_tuning.py
```diff
@@ -313,7 +313,11 @@ def export_model_preds(preds_file, predictions, validation_stream):
 
     # Init the resource & Build the tuning config from our dataset/arg info
     print_colored("[Loading the base model resource...]")
-    base_model = model_type.bootstrap(args.model_name, tokenizer_name=args.model_name)
+    base_model = model_type.bootstrap(
+        args.model_name, 
+        tokenizer_name=args.model_name, 
+        torch_dtype=args.torch_dtype
+    )
```
gkumbhat, line 320:
```suggestion
    base_model = model_type.bootstrap(
        args.model_name, tokenizer_name=args.model_name, torch_dtype=args.torch_dtype
    )
```

PR #143, opened 2023-08-23
Fix stream api gen

gkumbhat:
### Changes
- Remove wip tag from prompt tuning module for streaming function which causes the API generation to fail


caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -213,10 +213,10 @@ def run(
             )
         return GeneratedTextResult(generated_text=gen_text[0])
 
+    # @wip_decorator.work_in_progress(
```
gabe-l-hart, line 216: Why not just remove it outright?

gkumbhat: Super thanks for quick review. I just kept it to make a note that this function is "work in progress" and why we can't actually use the decorator.

gabe-l-hart (approved):
Ship it. Nice find!

Issue #140, opened 2023-08-22
Add extended inferencing parameter for text-generation inference 


gkumbhat:
## Is your feature request related to a problem? Please describe.

Currently TGIS exposes certain inference time parameters like temperature etc, which we haven't exposed in any of our text-generation modules. This can become limitation for users. This issue is to add those parameters to TGIS backend modules (both text-generation and prompt tuning) and also implement them to "local" backends

TGIS also exposes details about the input like token count in its response. These aren't exposed currently and should be added as well.

### Input params
- [ ] decoding_method (they call it method) 
- [ ] sampling_parameters (they call it sampling) 
- [ ] StoppingCriteria - `max_new_tokens`, `min_new_tokens` already there. `time_limit_millis` one is tricky for local implementation specially.
- [ ] DecodingParameters - This one is tricky for local implementation specially 
- [ ] `truncate_input_tokens` already supported in text_generation_* modules. Not available in peft ones.

#### Output params
- [ ] `input_token_count` - Need to add in `caikit.interfaces.nlp`

chakrn:
@gkumbhat can you close this out if already done?

gkumbhat:
Can be closed as already completed!

PR #151, opened 2023-08-28
✨ Text generation input inference data models

tharapalanivel:
Supports https://github.com/caikit/caikit-nlp/issues/140

### DIFF
--- a/caikit_nlp/data_model/generation.py
+++ b/caikit_nlp/data_model/generation.py
@@ -97,7 +97,7 @@ class SamplingParameters(DataObjectBase):
 @caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
 class StoppingCriteria(DataObjectBase):
 
-    max_tokens: int
-    min_tokens: int
+    max_new_tokens: int
+    min_new_tokens: int
     time_limit_millis: int
     stop_sequences: List[str]
--- a/tests/data_model/test_generation.py
+++ b/tests/data_model/test_generation.py
@@ -34,7 +34,7 @@ dummy_sampling_parameters = SamplingParameters(
 )
 
 dummy_stopping_criteria = StoppingCriteria(
-    max_tokens=200, min_tokens=50, time_limit_millis=0, stop_sequences=["Test"]
+    max_new_tokens=200, min_new_tokens=50, time_limit_millis=0, stop_sequences=["Test"]
 )
 
 ## Tests ########################################################################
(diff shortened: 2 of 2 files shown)


caikit_nlp/data_model/generation.py
```diff
@@ -71,3 +71,33 @@ class TuningConfig(DataObjectBase):
...
+    top_k: int
+    top_p: int
+    typical_p: float
+    seed: Optional[int]
+
+
+@caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
+class StoppingCriteria(DataObjectBase):
+
+    max_tokens: int
```
gkumbhat, line 100: this would be `max_new_tokens` and `min_new_tokens` right?

gkumbhat (changes requested):

caikit_nlp/data_model/generation.py
```diff
@@ -71,3 +71,33 @@ class TuningConfig(DataObjectBase):
...
+
+
+@caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
+class SamplingParameters(DataObjectBase):
+
+    temperature: float
+    top_k: int
+    top_p: int
+    typical_p: float
+    seed: Optional[int]
```
alex-jw-brooks, line 94: It might be a good idea to set some defaults here! TGIS defaults are [here](https://github.com/IBM/text-generation-inference/blob/main/router/src/lib.rs#L63).

Most of the time this doesn't matter, because 0 temperature (in the IBM fork) indicates greedy decoding, so `top_k`, `top_p`, `typical_p`, etc won't be used, as they're sampling only.

TGI doesn't use temperature 0 as a toggle though, so it would be also be nice in case those APIs are ever more unified - currently there is some small divergences with stuff like prompt IDs

alex-jw-brooks (changes requested):

caikit_nlp/data_model/generation.py
gkumbhat: I think even if we set defaults on the DM, they won't propagate to proto, so the default here would be guided by the `.run` function themselves.

alex-jw-brooks: Good point - my main concern with leaving it up to run is that it's easy for defaults to get out of sync if we have multiple modules relying on them. 

I guess an alternate is to either have a building for getting these objects with their default values that make sense, or to have consts be passed to the run function 🤔 is the intent with this type to have a parameter that is this DM object type, or to take primitives and build this object in the requests?

gkumbhat:
We decided to go with flattened API for now. This may be revisited, so we are keeping this PR open

PR #152, opened 2023-08-29
Flattened inf params

gkumbhat:
Supports #155

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
+++ b/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
@@ -65,6 +65,7 @@ from ...resources.pretrained_model import (
     HFAutoSeq2SeqLM,
     PretrainedModelBase,
 )
+from ...toolkit.text_generation.model_run_utils import GENERATE_FUNCTION_ARGS, generate_text_func
 from ...toolkit.data_stream_wrapper import SimpleIterableStreamWrapper
 from ...toolkit.data_type_utils import get_torch_dtype, str_to_torch_dtype
 from ...toolkit.task_specific_utils import convert_to_generation_record
@@ -170,7 +171,6 @@ class PeftPromptTuning(ModuleBase):
     def run(
         self,
         text: str,
-        device: Optional[Union[str, int]] = _DETECT_DEVICE,
         max_new_tokens: Optional[int] = 20,
         min_new_tokens: Optional[int] = 0,
         truncate_input_tokens: Optional[int] = 0,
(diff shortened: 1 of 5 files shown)


caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -170,48 +174,51 @@ def __del__(self):
     def run(
         self,
         text: str,
-        device: Optional[Union[str, int]] = _DETECT_DEVICE,
-        max_new_tokens=20,
-        min_new_tokens=0,
+        max_new_tokens: Optional[int] = 20,
+        min_new_tokens: Optional[int] = 0,
+        truncate_input_tokens: Optional[int] = 0,
+        decoding_method: Optional[str] = "GREEDY",
```
alex-jw-brooks, line 185: Any thoughts on API alignment for the modules between this and the TGIS remote? Specifically, the TGIS API doesn't use `decoding_method`. Having the two modules diverge, even if transformers/TGIS do, feels a bit strange since the args really are doing the same thing 🤔

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -170,48 +174,51 @@ def __del__(self):
...
+        top_k: Optional[int] = 0,
+        top_p: Optional[float] = 0.0,
+        typical_p: Optional[float] = 0.0,
+        temperature: Optional[float] = 1.0,
+        repetition_penalty: Optional[float] = 0.0,
```
alex-jw-brooks, line 191: I think the default repetition penalty should be `1.0`! 1.0 means no penalty, and is the default for both the login processor in transformers and TGIS

https://huggingface.co/docs/transformers/main_classes/text_generation#transformers.GenerationConfig.repetition_penalty

https://github.com/IBM/text-generation-inference/blob/main/router/src/lib.rs#L80

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -170,48 +174,51 @@ def __del__(self):
...
         text: str,
-        device: Optional[Union[str, int]] = _DETECT_DEVICE,
-        max_new_tokens=20,
-        min_new_tokens=0,
+        max_new_tokens: Optional[int] = 20,
+        min_new_tokens: Optional[int] = 0,
+        truncate_input_tokens: Optional[int] = 0,
+        decoding_method: Optional[str] = "GREEDY",
+        top_k: Optional[int] = 0,
+        top_p: Optional[float] = 0.0,
```
alex-jw-brooks, line 187: Default `top_p` should also probably be `1.0`, I believe. Otherwise, leaving this unset and setting sampling params will default to greedy decoding

This is also what TGIS does - `1.0` essentially means consider everything: https://github.com/IBM/text-generation-inference/blob/main/router/src/lib.rs#L70C1-L72C8

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -170,48 +174,51 @@ def __del__(self):
...
+            truncate_input_tokens=truncate_input_tokens,
+            decoding_method=decoding_method,
+            top_k=top_k,
+            top_p=top_p,
+            typical_p=typical_p,
+            temperature=temperature,
+            repetition_penalty=repetition_penalty,
+            max_time=max_time,
+            exponential_decay_length_penalty=exponential_decay_length_penalty,
+            stop_sequences=stop_sequences,
```
alex-jw-brooks, line 229: What is the motivation for always passing the values here and setting our own defaults, rather than setting ours to None by default in `.run`, and only overriding the `not None` args (i.e., pulling the defaults from .generate instead)?

caikit_nlp/modules/text_generation/text_generation_local.py
```diff
@@ -380,106 +384,51 @@ def save(self, model_path):
...
-        truncate_input_tokens: int = 0,
+        max_new_tokens: Optional[int] = 20,
+        min_new_tokens: Optional[int] = 0,
+        truncate_input_tokens: Optional[int] = 0,
+        decoding_method: Optional[str] = "GREEDY",
+        top_k: Optional[int] = 0,
+        top_p: Optional[float] = 0.0,
+        typical_p: Optional[float] = 0.0,
+        temperature: Optional[float] = 1.0,
+        repetition_penalty: Optional[float] = 0.0,
```
alex-jw-brooks, line 395: Same comments for repetition penalty and top_p here

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -0,0 +1,241 @@
...
+        Options: "GREEDY" (default), "SAMPLING"
+    top_k: int
+        The number of highest probability vocabulary tokens to keep for
+        top-k-filtering. Only applicable when decoding_method is SAMPLING.
+        Default: 0 - means disabled
+    top_p: float
+        If set to float < 1, only the smallest set of most probable tokens
+        with probabilities that add up to top_p or higher are kept for
+        generation. Only applicable when decoding_method is SAMPLING.
+        Default: 0.0 - means disabled - equivalent to 1.0
```
alex-jw-brooks, line 65: Is there a source for this? I took a quick glance at the top p logit warper, and don't see 0 being handled as a special case, although I haven't looked closely into how this hooks into generate very recently 

Top P logic warper [source](https://github.com/huggingface/transformers/blob/main/src/transformers/generation/logits_process.py#L388) for reference, very possible I am missing something

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -0,0 +1,241 @@
...
+    if truncate_input_tokens == 0:
+        # NOTE: below will make model throw error in case inputs are longer
+        # than allowed length
+        truncation = False
+
+    else:
+        truncation = True
+
+    if repetition_penalty == 0.0:
+        repetition_penalty = 1.0
```
alex-jw-brooks, line 201: Ah, I see now - but why is this being overridden here rather than just using 1 as the default in .run?

Is top_p supposed to be handled outside of generate like this too?

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+
+    # NOTE: below is to match TGIS API, where 0 identifies as no truncation
+    if truncate_input_tokens == 0:
+        # NOTE: below will make model throw error in case inputs are longer
+        # than allowed length
+        truncation = False
+
+    else:
+        truncation = True
```
alex-jw-brooks, line 198: nit: could use `truncation = truncate_input_tokens != 0` to shorten this a bit

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -0,0 +1,241 @@
...
+        int,
+        allow_none=True,
+        truncate_input_tokens=truncate_input_tokens,
+    )
+    error.type_check("<NLP84635843E>", int, allow_none=True, top_k=top_k)
+    error.type_check("<NLP55267523E>", float, allow_none=True, top_p=top_p)
+    error.type_check("<NLP13670202E>", float, allow_none=True, typical_p=typical_p)
+    error.type_check(
+        "<NLP11929418E>", float, allow_none=True, repetition_penalty=repetition_penalty
+    )
```
alex-jw-brooks, line 179: Are we leaving generate to do all the validation for values, or would it be better to explicitly check that the max tokens >= min tokens?

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -0,0 +1,241 @@
...
+
+    preds = [
+        tokenizer.decode(g, skip_special_tokens=True, clean_up_tokenization_spaces=True)
+        for g in generate_ids
+    ]
+
+    if generate_ids[0][-1].item() == eos_token:
+        finish_reason = "EOS_TOKEN"
+    elif generate_ids.size(1) - 1 == max_new_tokens:
+        finish_reason = "MAX_TOKENS"
```
alex-jw-brooks, line 250: Definitely a technicality, but I'm wondering if MAX_TOKENS should be the reason if we hit the maximum number of tokens and the last one is an EOS 🤔 in my opinion, this makes a little more sense

tests/modules/text_generation/test_peft_prompt_tuning.py
```diff
@@ -297,3 +299,48 @@ def test_model_can_only_have_one_or_two_transformer_modules(seq2seq_lm_dummy_mod
...
+    """
+    pred = causal_lm_dummy_model.run(
+        "This text doesn't matter",
+        decoding_method="GREEDY",
+        top_k=2,
+        top_p=0.23,
+        typical_p=0.23,
+        temperature=0.77,
+    )
+    assert isinstance(pred, GeneratedTextResult)
```
alex-jw-brooks, line 335: It might be nice to add some kind of validation to trace the input args to generate or something like that? Since these tests aren't actually validating that greedy / sampling etc are happening

alex-jw-brooks (commented):
This looks great! Some initial thoughts, but mostly just discussion points

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: Yep, We would absolutely add support for TGIS modules for all these. That work is going on in parallel by @tharapalanivel .

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: 🤔  That make sense. 1.0 and 0.0 are equivalent, so we'll change this

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: In some places it was to align with TGIS (as per the proto def)

caikit_nlp/toolkit/text_generation/model_run_utils.py
gkumbhat: good point. Will add this validation in here as well

caikit_nlp/toolkit/text_generation/model_run_utils.py
gkumbhat: This was to align with tgis, as per the comments in proto. https://github.com/caikit/caikit-tgis-backend/blob/main/caikit_tgis_backend/generation.proto#L103

caikit_nlp/toolkit/text_generation/model_run_utils.py
gkumbhat: hmm. just to make sure I am understanding correctly. I guess, that would depend if the last token would still be EOS in case we hit the size 🤔

caikit_nlp/toolkit/text_generation/model_run_utils.py
gkumbhat: this was mentioned in the proto file

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
alex-jw-brooks: Cool sounds good. Also might be good to think about whether or not we support TGI somehow - it's a bit weird when the proto specs are a bit different

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
alex-jw-brooks: It's only equivalent in TGIS! Not in the transformers library

caikit_nlp/toolkit/text_generation/model_run_utils.py
alex-jw-brooks: No idea! I think I can check that and get back to you 😄

Let's not let that block this PR though, can submit a patch if it's not aligned later, since this is a super obscure case and both make sense

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
alex-jw-brooks: Makes sense

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat: That is great to know. 🤔

gkumbhat: I agree, its probably good to align with defaults in transformers. The good thing is for `top_p` and `typical_p`, the default in transformers is also `1.0`. So the only one that is different is `top_k` which we have as 0 but transformers has as 50

Issue #167, opened 2023-09-02
Text generation API field marked with incorrect type


gkumbhat:
## Describe the bug

1. `stop_sequences` parameter in text generation module is supposed to be `List[str]` but it is marked as `str` [here](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L189). 
2. `id` field in `GeneratedToken` datamodel is not getting used. This can cause confusion to the users.

PR #171, opened 2023-09-05
Fix seed type mismatch

tharapalanivel:
Supports #167


caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -180,7 +181,7 @@ def run(
         top_p: Optional[float] = 1.0,
         typical_p: Optional[float] = 1.0,
         temperature: Optional[float] = 1.0,
-        seed: Optional[int] = None,
+        seed: Optional[np.uint64] = None,
```
gkumbhat, line 184: hmm. curious, why does it need to be `np.uint64` vs int ?

gkumbhat: make sense. Thanks!

gkumbhat (approved):
LGTM

PR #173, opened 2023-09-05
:truck: Change parameter name and mark them optional for runtime

gkumbhat:
### Changes
- Change `lr` to `learning_rate` to make it more intuitive 
- Mark training parameter with default as optional via type annotation for the correct runtime service generation.


alex-jw-brooks (approved):
Looks good to me - aside from being more correct type hints, does the lack of optionals here break runtime train jobs if they're not set? I need to learn more about the way type hints are leveraged by it 😄

gkumbhat:
Thanks @alex-jw-brooks . It doesn't break, just doesn't show up as optional so adds confusion

PR #174, opened 2023-09-05
[caikit-nlp-163] Refactor peft module to take out common peft config functionality

rawkintrevo:


rawkintrevo:
@gkumbhat i moved a batch of stuff to a new file- is this what you had in mind? wanted to confirm i was on the right path before i did more stuff

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
gkumbhat, line 1: Thanks @rawkintrevo for quick turn-around. I think 1 other thing that would move would be https://github.com/caikit/caikit-nlp/pull/174/files#diff-1cb191003903163320c02f8ffaf7c5edd48ca6649cd3092d1b4c3a0fdcd038c1R355 function, since that returns the  `peft_config` which is then passed to get the `peft_model`

rawkintrevo: do you meant this line? https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L519

caikit_nlp/modules/text_generation/peft_config.py
```diff
@@ -0,0 +1,193 @@
...
+    PROMPT_TUNING = "PROMPT_TUNING"
+    MULTITASK_PROMPT_TUNING = "MULTITASK_PROMPT_TUNING"
+    # MULTITASK_PREFIX_TUNING = "MULTITASK_PREFIX_TUNING"
+    # P_TUNING = "P_TUNING"
+    # PREFIX_TUNING = "PREFIX_TUNING"
+    # LORA = "LORA"
+
+
+def validate_peft_config(
+    tuning_type, tuning_config, error, log, base_model, cls, torch_dtype, verbalizer
```
gkumbhat, line 53: We usually initialize logger per module / file to allow easy backtracking.  To do this, we can initialize `log` at the beginning of this file and re-use that everywhere in this file.
```
log = alog.use_channel("PFT_CNFG_TLKT")
```

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -362,133 +340,26 @@ def train(
...
+        metric = kwargs.get("metric")
 
-                error.file_check(
-                    "<NLP96030210E>", tuning_config.prompt_tuning_init_source_model
-                )
-                log.debug(
-                    "Validated tuning source prompt [%s]",
-                    tuning_config.prompt_tuning_init_source_model,
-                )
+        base_model_name = base_model._model_name
```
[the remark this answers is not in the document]
rawkintrevo: The linked block exists in peft_config.py at line 100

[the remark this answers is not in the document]
gkumbhat: yep, but the problem is that one is getting called after we do line 347. And so line 347 can fail if `base_model` is a string. 

Since the code to resolve the base_model is common. It might make sense to pull that out from `validate_peft_config` function into a separate `resolve_base_model` function and we call that function before we call `validate_peft_config`.. So essentially:
```
base_model = resolve_base_model(base_model)
task_type, output_model_types, peft_config, tuning_type = validate_peft_config(..., base_model, ...)
```

gkumbhat (changes requested):

gkumbhat (approved):
LGTM. Thanks Trevor. We need to coordinate these merges a bit as they are all changing same files.

rawkintrevo:
ack @gkumbhat - will let someone else coordinate the merge

alex-jw-brooks:
Pushed a rebased version of this branch to https://github.com/caikit/caikit-nlp/pull/197!

PR #185, opened 2023-09-11
:bug: Fix attention mask and task_id for MPT

gkumbhat:
### Changes
- For causal-lm models we need attention mask to be able to run generate function
- `task_ids` is now a compulsory field in PEFT. However, we won't be getting this from users. Since we do not allow task specific selection at run time, and use methods like average, we can just fill the task ids with 0 vector to fix this issue.


alex-jw-brooks (approved):
Thanks Gaurav. Looks good!

PR #193, opened 2023-09-16
fix: use cli args for metrics to run

timrbula:
# Changes

- update arg for `metric` to use `choices` param to check the arg passed
- remove manual check for arg passed
- fix `metric_funcs` to be based on arg

### DIFF
--- a/examples/evaluate_model.py
+++ b/examples/evaluate_model.py
@@ -136,8 +136,9 @@ def export_model_preds(preds_file, predictions, validation_stream, verbalizer):
 if __name__ == "__main__":
     configure_random_seed_and_logging()
     args = parse_args()
-    print_colored("Metrics to be calculated: {}".format(args.metrics))
     metric_funcs = [SUPPORTED_METRICS[metric] for metric in args.metrics]
+    print_colored("Metrics to be calculated: {}".format(args.metrics))
+
     # Load the model; this can be a local model, or a distributed TGIS instance
     print_colored("Loading the model...")
     model = load_model(args.tgis, str(args.model_path))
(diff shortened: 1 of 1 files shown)


examples/evaluate_model.py
```diff
@@ -135,14 +136,9 @@ def export_model_preds(preds_file, predictions, validation_stream, verbalizer):
...
     configure_random_seed_and_logging()
     args = parse_args()
-    if not all(metric_name in SUPPORTED_METRICS for metric_name in args.metrics):
-        raise KeyError(
-            "One or more requested metrics {} not supported! Supported metrics: {}".format(
-                args.metrics, list(SUPPORTED_METRICS.keys())
-            )
-        )
-    metric_funcs = list(SUPPORTED_METRICS.values())
+    metric_funcs = [SUPPORTED_METRICS[metric] for metric in args.metrics]
```
gkumbhat, line 139: nit: Lets add the `keyerror` validation back, in case there is a typo or something when kicking off evaluation.

gkumbhat (changes requested):
Left a small nit, other than that looks good!

examples/evaluate_model.py
timrbula: My idea here was for it to checked by argparse via 'options' so if you make a typo it will fail at the parseargs call. Do you want an additional level of checking?

gkumbhat: Oh yes, that make sense. Sounds good..

gkumbhat (approved):
Thanks for the fix @timrbula .  LGTM

gkumbhat:
@timrbula Can you amend your commit with `--signoff` to get the DCO check pass

PR #201, opened 2023-09-21
Expose seed

tharapalanivel:
For reproducibility purposes, the random seed used for inference should be returned to user.


gkumbhat (approved):
LGTM

PR #206, opened 2023-09-25
Causal LM tokenization: Chunking and seq2seq Forward

alex-jw-brooks:
This PR is the nth rewrite of causal language modeling tokenization, _now_ changing it to:
- Let us use sequence type tokenization with causal language models if we want to
- Or enable chunking on the concatenated sequence, optionally dropping the remainder if the chunks are uneven

Sequence type tokenization is a port of the old causal LM tokenization logic with a bug fix for the concatenation sequence length; I have verified the quality matches approximately what we would expect from prompt tuning on RTE.

Chunking seems to be functional and has tests verifying output correctness but is not publicly exposed yet and therefore has not fully quality tested.

We need to be careful about collator compatibility when changing the chunking strategy, since prompt tuning and the base model resources/text generation currently manage collators for different model types differently. The main thing to be wary of is that the causal language modeling collator should _not_ be used with the sequence approach to avoid clobbering the labels.

Note that currently things are still being statically padded and using the default collator; it would be more ideal to:
- use the padding collator for dynamic padding when possible
- add better collator based tests operating on transformers datasets; this PR does start these tests a bit, but only covers the default collator, which does not necessarily need a torch dataset to work

### DIFF
--- a/caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
+++ b/caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
@@ -66,7 +66,7 @@ class HFAutoCausalLM(PretrainedModelBase):
         max_target_length: int,
         verbalizer: Union[None, str] = None,
         task_ids: Union[None, int] = None,
-        use_seq2seq_tokenization: bool = False,
+        use_seq2seq_tokenization: bool = True,
         chunk_size: int = 128,
         drop_remainder: bool = False,
     ) -> DataStream[BatchEncoding]:
--- a/caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
+++ b/caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
@@ -19,6 +19,7 @@ from collections.abc import Mapping
 from typing import List, Union
 
 # Third Party
+import torch
 from transformers import (
     AutoModelForCausalLM,
     BatchEncoding,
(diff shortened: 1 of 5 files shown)


caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
```diff
@@ -177,13 +178,84 @@ def _get_data_collator(self, **kwargs) -> "transformers.DataCollator":
...
+            Union[DataStream[BatchEncoding], BatchEncoding]
+                Encoded chunked sequence as a stream or batch encoding object.
+        """
+        source_ids = tokenizer(source, max_length=max_source_length, truncation=True)
+        target_ids = tokenizer(target, max_length=max_target_length, truncation=True)
+
+        # Force everything to a list of batch encodings; for non-batch mode, this just
+        # puts it into a list. For batch mode, we get a list of batch encodings,
+        # allowing us to standardize subsequent processing a bit.
+        source_id_chunks = cls._force_to_batch_encoding_list_of_chunks(
```
gkumbhat, line 233: nit: for readability, can we add an example in comments of how the `source_id_chunks` would look like

caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
```diff
@@ -198,69 +270,123 @@ def _force_to_batch_encoding_list(
...
-            num_target_samples: int
-                Length of the original target seequence being considered.
+            left: BatchEncoding
+                Encoding representing left sequence, which will be updated in place.
+                Corresponds to source.
+            right: BatchEncoding
+                Encoding representing right sequence, which will be stacked onto the left
+                encoding. Corresponds to target.
+        """
+        for k in left.keys():
```
gkumbhat, line 332: since this will get called a lot of time, shall we use dictionary comprehension here?

gkumbhat (approved):
LGTM. Thanks Alex. The default for now for seq2seq make sense. Lets get the other changes validated and working iteratively and then enable that option. For now, getting this in to unblock causal-lm prompt tuning.

PR #208, opened 2023-09-27
Add ft loss logging

gkumbhat:
### Changes
- Update loss logging mechanism to log per step.
    - NOTE: The definition of step changes depending on how the training logic is implemented and if we are using `gradient_accumulation`. In case of huggingface trainer with gradient accumulation greater than 1, the definition of step becomes when the gradients are updated, which would be at every gradient accumulation step. 
- Update prompt tuning to use step wise gradient logic
- Add utility function for trainers to use for logging 
- Extend Trainer class for both base resource and seq2seq resource to override the `log` function
- Add loss logging functionality for fine-tuning module, similar to prompt tuning.

### Test

Produces following logs in `training_logs.jsonl` for fine-tuning:
```
{"name": "loss", "data": {"epoch": 0.01, "step": 1, "value": 3.9274, "timestamp": "2023-09-26T19:25:18.737364"}}
{"name": "loss", "data": {"epoch": 0.01, "step": 2, "value": 4.4904, "timestamp": "2023-09-26T19:25:19.316729"}}
...
```

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
+++ b/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
@@ -1042,7 +1042,7 @@ class PeftPromptTuning(ModuleBase):
 
         training_loss_tracker = []
 
-        step_count = 0
+        step_count = 1
 
         for epoch in range(num_epochs):
             step_loss_log = {}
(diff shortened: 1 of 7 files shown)


caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -1060,21 +1063,27 @@ def _execute_train_loop(
...
-                }
-            )
+
+            for step, loss_val in step_loss_log.items():
+
+                # Below is added to be propagated and stored as training_metadata
+                training_loss_tracker.append(
+                    {
+                        "epoch": epoch,
+                        "step": step,
```
gkumbhat, line 1082: This is to start logging step wise instead of per epoch. This will change a bit with Trainer, but the general format of logging would remain same.

alex-jw-brooks (approved):
looks good - thanks!

PR #210, opened 2023-09-28
return correct finish reason in generate_text_func

dtrifiro:
The finish reason in `generate_text_func` is currently broken (always returns `OTHER`). The reason is that a comparison is being made between a tensor element (`int)` and a string (`eos_token`) without converting it using the `tokenizer`

This PR fixes the tokenization issue and adds a few type hints

Full reproduction screiopt below

## Reproduction script
```bash
#!/bin/sh
set -exu

tempdir="$(mktemp -d)"
cd ${tempdir}

python -m venv .venv 
source .venv/bin/activate
pip install git+https://github.com/caikit/caikit-nlp

grpcurl="${tempdir}/grpcurl"

# this requires git-lfs to be installed and `git lfs install` to have been run
if ! command -v git-lfs &>/dev/null; then echo "this requires git-lfs"; fi

git clone https://huggingface.co/google/flan-t5-small

# convert the model for caikit
python -c 'import caikit_nlp; model = caikit_nlp.text_generation.TextGeneration.bootstrap("flan-t5-small"); model.save("flan-t5-small-caikit")'

CAIKIT_LOG_FILE=$(mktemp)

export TRANSFORMERS_CACHE="$PWD"
export RUNTIME_LIBRARY=caikit_nlp
export RUNTIME_LOCAL_MODELS_DIR="$PWD"

python -m caikit.runtime.grpc_server &>$CAIKIT_LOG_FILE &

# wait for caikit to properly start
timeout --signal=SIGINT 30 bash -c "grep -m1 'Caikit Runtime is serving on port' <(tail -f $CAIKIT_LOG_FILE)"

cat $CAIKIT_LOG_FILE

# install grpcurl to actually query the endpoint
curl -sL https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz | tar zxvf - grpcurl
./grpcurl -plaintext -d '{"text": "At what temperature does liquid Nitrogen boil?"}' -H "mm-model-id: flan-t5-small-caikit" localhost:8085 caikit.runtime.Nlp.NlpService/TextGenerationTaskPredict
rc=$?

kill %1
exit $c
```


dtrifiro:
Tagging @Xaenalt for visibility

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -233,7 +238,9 @@ def generate_text_func(
         for g in generate_ids
     ]
 
-    if generate_ids[0][-1].item() == eos_token:
+    if (eos_token and tokenizer.decode(generate_ids[0, -1].item()) == eos_token) or (
```
dtrifiro, line 241: Also adding an extra check for the `eos_token`, since it's actually an `Optional[str]`, falling back to `tokenizer.eos_token_id` if it is not defined. Hopefully this is the correct behaviour.

dtrifiro:
A small sidenote here:

https://github.com/caikit/caikit-nlp/blob/ef283e47485775d544ece31507f4a39bd994663d/caikit_nlp/toolkit/text_generation/model_run_utils.py#L236-L241

If `finish_reason` is set to  `OTHER", `caikit` currently fails to serialize the response because that's not defined in the `FinishReason` enum in caikit:
https://github.com/caikit/caikit/blob/4d01c9b4caded5492ca5772875f0d8778143a657/caikit/interfaces/nlp/data_model/text_generation.py#L35-L44

Would it make sense to add it there?

Xaenalt:
> A small sidenote here:
> 
> https://github.com/caikit/caikit-nlp/blob/ef283e47485775d544ece31507f4a39bd994663d/caikit_nlp/toolkit/text_generation/model_run_utils.py#L236-L241
> 
> If `finish_reason` is set to `OTHER", `caikit`currently fails to serialize the response because that's not defined in the`FinishReason` enum in caikit: https://github.com/caikit/caikit/blob/4d01c9b4caded5492ca5772875f0d8778143a657/caikit/interfaces/nlp/data_model/text_generation.py#L35-L44
> 
> Would it make sense to add it there?

I think it does make sense to add it there, especially if it's being referenced here. @gabe-l-hart is this intended?

gabe-l-hart:
Hm, that does seem troubling. @gkumbhat I'll defer to you on this, but I think we probably do need to add `OTHER` to the enum in `caikit/interfaces`

gkumbhat (commented):
Just left a small comment, but otherwise looks good. Thanks for catching this and contributing the fix.

gkumbhat:
Actually `OTHER` here is representing a lot of possible scenarios. We need to map it back to proper reasons I think. The enum we have in caikit is aligned with HF and with TGIS [here](https://github.com/caikit/caikit-tgis-backend/blob/main/caikit_tgis_backend/generation.proto#L151). Adding `OTHER` in `caikit` interfaces would cause confusion and would not be informative. 🤔

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -36,6 +36,11 @@
 # Local
 from ...data_model import ExponentialDecayLengthPenalty
 
+if TYPE_CHECKING:
+    # Third Party
+    from transformers import AutoModel, AutoTokenizer
```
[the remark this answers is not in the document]
dtrifiro: Not sure what you mean here. Get rid of `if TYPE_CHECKING`? `transformers` is already a dependency in `pyproject.toml`

dtrifiro:
Thinking about it some more, the only other finish reason that should be returned by this function is `STOP_SEQUENCE`. By quickly glancing at the code, I'm not sure this scenario works right now. I propose opening a new issue and tackle this in other PR. I think the overall objective would be to get rid of `OTHER` here since it breaks the protocol anyway.

caikit_nlp/toolkit/text_generation/model_run_utils.py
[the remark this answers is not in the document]
gkumbhat: @dtrifiro yep. thats exactly what I was suggesting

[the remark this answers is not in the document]
dtrifiro: Done

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -233,10 +233,19 @@ def generate_text_func(
...
     elif generate_ids.size(1) - 1 == max_new_tokens:
         finish_reason = "MAX_TOKENS"
+    elif ("stopping_criteria" in gen_optional_params) and (
+        gen_optional_params["stopping_criteria"](
+            generate_ids,
+            None,  # scores, unused by SequenceStoppingCriteria
+        )
+    ):
+        finish_reason = "STOP_SEQUENCE"
     else:
```
gkumbhat, line 249: Lets remove `OTHER` since it creates issues anyways?

dtrifiro: What do you propose to do in that case? I'd rather have some way of handling the else branch (`RuntimeError`? )

gkumbhat: Yeah, we can do `error("<>", ...)` that way its evident something went wrong

dtrifiro: What error code should be using though? I think the exception should be something in the lines of `RuntimeError("Could not determine finish reason")`

gkumbhat: Yep. An example of throwing this via `error` : 

https://github.com/caikit/caikit/blob/main/caikit/core/data_model/streams/data_stream.py#L247


The error code over there, i.e `<COR59442457E>` is of the format: `<3 char lib identifier><8 int random number>E`

1. `<3 char lib identifier>` for caikit nlp is `NLP`
2. E suffix identifies this as a ERROR level message

markstur: @gkumbhat I was wondering where those error numbers come from.  So something like this to generate on mac is okay?

```ShellSession
markstur@Marks-MacBook-Pro-3 ~ % echo $RANDOM$RANDOM
439318353
```

And whatever padding/truncation you like to grab 8 random-looking digits?  Wierd... but that should help search code.

dtrifiro: @gkumbhat Thanks. Updated the code with `error("<NLP38342991E>", RuntimeError(...))`

@markstur 
In python: 
```python
from random import choices

lib_identifier = "NLP"
message_type = "E"
unique_code = "".join(map(str, choices(range(10), k=8)))

print(f"{lib_identifier}{unique_code}{message_type}")
```

dtrifiro:
@gkumbhat Another fix which was required in order for CI to pass with the new exception instead of `OTHER` as finish reason, is to fix the `finish_reason` logic for `MAX_TOKENS`, since it was currently broken causing most tests to break. See the last commit

caikit_nlp/toolkit/text_generation/model_run_utils.py
gkumbhat: If you use vscode, you can use this extension to automatically fill appropriate log code: https://marketplace.visualstudio.com/items?itemName=Gaurav-Kumbhat.alog-code-generator

caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -237,7 +237,9 @@ def generate_text_func(
         generate_ids[0, -1] == tokenizer.eos_token_id
     ):
         finish_reason = "EOS_TOKEN"
-    elif generate_ids.size(1) - 1 == max_new_tokens:
+    elif (generate_ids.size(1) - 1 == max_new_tokens) or (
+        generate_ids.size(1) - inputs["input_ids"].size(1) == max_new_tokens
```
gkumbhat, line 241: hmm. technically input will only be in output for `causal-lm` type of models, so there can be some side-effects in this calculations  🤔  May be we would need to check if input is actually in output and then calculate max_new_tokens for that accordingly

dtrifiro: @gkumbhat  I agree. I gave some thought to it and I ended up with the logic in the latest push. The idea is the following:

1. Check if the last token is `eos_token`, if so, `finish_reason = "EOS_TOKEN"`
2. Check if the stop sequence is the the generated tokens, if so, `finish_reason = "STOP_SEQUENCE"`
3. If none of the above conditions are true, then `finish_reason = "MAX_TOKENS"`

My reasoning for always returning `MAX_TOKENS` comes from looking at [`caikit.interfaces.nlp.data_model.text_generation.FinishReason`]( https://github.com/caikit/caikit/blob/4d01c9b4caded5492ca5772875f0d8778143a657/caikit/interfaces/nlp/data_model/text_generation.py#L35-L44):

Ia added some comments to each enum value
```python
@dataobject(package=NLP_PACKAGE)
class FinishReason(Enum):
    NOT_FINISHED = 0 # should only be returned by streaming implementations
    MAX_TOKENS = 1  
    EOS_TOKEN = 2  # matches rule #1 above
    CANCELLED = 3  # should not be set by generate_text_func
    TIME_LIMIT = 4 # should not be set by generate_text_func
    STOP_SEQUENCE = 5 # matches rule #2 above
    TOKEN_LIMIT = 6 # not sure about this
    ERROR = 7  # fairly generic, 
```

I'm unsure about `TOKEN_LIMIT` (how is this different from `MAX_TOKENS`?), but I think none of the other enum values apply to `generate_text_func` .

gkumbhat (approved):
LGTM

caikit_nlp/toolkit/text_generation/model_run_utils.py
gkumbhat: `TOKEN_LIMIT`refers to the maximum number of tokens limit defined by the model whereas the `MAX_TOKENS` refers to the maximum number defined by the user. So one can reach `TOKEN_LIMIT` before `MAX_TOKENS`

dtrifiro: @gkumbhat Ok, so that means that also need to check whether we reached the token limit for the model, although I'm not sure how we can get that information. Can we open an issue for adding support for `TOKEN_LIMIT` so that we can work on that on another PR?

dtrifiro: Created https://github.com/caikit/caikit-nlp/issues/253

dtrifiro:
Hey @gkumbhat, is there anything I can do to get this merged?

PR #211, opened 2023-09-28
:sparkles: Add support for reading base model from base models dir

gkumbhat:
### Changes
- Add support for optionally configuring the base_model base path via config or environment variable while loading the base model for prompt tuning


caikit_nlp/modules/text_generation/peft_config.py
```diff
@@ -56,6 +58,16 @@ class TuningType(str, Enum):
 
 def resolve_base_model(base_model, cls, torch_dtype):
     if isinstance(base_model, str):
+
+        if get_config().base_models_dir:
+
+            error.value_check(
```
evaline-ju, line 64: as written to validate the `base_model` shouldn't this be applicable whether or not there is a `base_models_dir`? or was the check meant to validate the directory itself?

gkumbhat: actually this case was meant to check for the case where we are considering to look at it from directory perspective, otherwise we leave it to the transformers library. but there is a miss here in terms of implementation. Good catch

evaline-ju: update makes more sense now! will wait on unit test additions and a format fix

tests/modules/text_generation/test_peft_config.py
```diff
@@ -63,3 +70,26 @@ def test_get_peft_config(train_kwargs, dummy_model, request):
     assert peft_config.task_type == dummy_resource.TASK_TYPE
     assert peft_config.prompt_tuning_init == tuning_config.prompt_tuning_init_method
     assert peft_config.prompt_tuning_init_text == tuning_config.prompt_tuning_init_text
+
```
evaline-ju, line 73: q - is there a test case where any of the `if get_config().base_models_dir` block gets tested since that's empty by default in the config?

evaline-ju (approved):
LGTM - thanks!

Issue #163, opened 2023-08-31
Refactor peft module to take out common peft config functionality


gkumbhat:
## Description

In order to re-use peft configuration in fine-tuning modules, we want to take out the common peft configuration logic to [toolkit/text_generation](https://github.com/caikit/caikit-nlp/tree/main/caikit_nlp/toolkit/text_generation) so that we can reuse these with couple of function calls from both of our tuning functions.

This includes, move of:
1. `allowed_tuning_init_methods`
2. `TuningType`
3. peft config creation and its usage [here](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L519)

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

chakrn:
@alex-jw-brooks says I can close this

alex-jw-brooks:
Link to rebase PR that was merged: https://github.com/caikit/caikit-nlp/pull/197

Issue #169, opened 2023-09-05
Add LoRA configuration support for PEFT module


gkumbhat:
## Description

After refactoring #163 PT to take out prompt tuning related configuration to separate module. We want to add support for prompt tuning (specifically LoRA) in fine-tuning module, i.e [text_generation](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/text_generation_local.py#L140). Changes:
1. Add LoRA configuration support in the toolkit functionality created in #163 
2. Add LoRA configuration and training from text-generation module (train).
3. Expose parameter required for LoRA configuration via `.train` function
4. Add support for saving LoRA models with "merged weights". This is to be done in `.train` function itself, that way, the `model` that we configure to `__init__` function will look like any other transformers model.

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] README updated [here](https://github.com/caikit/caikit-nlp#:~:text=Module(s)-,Salient%20Feature(s),-Text%20Generation)
- [ ] Example for training lora added in [example script](https://github.com/caikit/caikit-nlp/blob/main/examples/run_fine_tuning.py)

ibm-peach-fish:
@gkumbhat please confirm that for description items 2,3,4 you want https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/text_generation_local.py#L172 updated only ( I didn't see a `train` function in `text_generation_tgis.py` )

ibm-peach-fish:
@gkumbhat re: point 2. 
> Add LoRA configuration and training from text-generation module (train).
is `from` a typo that should be `to`?

ibm-peach-fish:
From slack conversation- add flag about weather model is a lora or not to the config.yaml

ibm-peach-fish:
also per slack conversation- FYI if anyone is doing something later and happens to read this (good for you). 

Merged models have lower latency than base + LoRA 

src: https://huggingface.co/docs/peft/conceptual_guides/lora

PR #212, opened 2023-09-28
[caikitnlp-169] Add LoRA configuration support for fine-tuning module

ibm-peach-fish:
- [x] Add LoRA configuration support in the toolkit functionality created in https://github.com/caikit/caikit-nlp/issues/163
- [x] Add LoRA configuration and training from text-generation module (train).
- [x] Expose parameter required for LoRA configuration via .train function
- [x] Add support for saving LoRA models with "merged weights". This is to be done in .train function itself, that way, the model that we configure to `__init__` function will look like any other transformers model.

- [x] Unit tests cover new/changed code
- [x] Examples build against new/changed code
- [ ] README updated [here](https://github.com/caikit/caikit-nlp#:~:text=Module(s)-,Salient%20Feature(s),-Text%20Generation)
- [x] Example for training lora added in [example script](https://github.com/caikit/caikit-nlp/blob/main/examples/run_fine_tuning.py)

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_config.py
+++ b/caikit_nlp/modules/text_generation/peft_config.py
@@ -17,7 +17,7 @@ from enum import Enum
 import os
 
 # Third Party
-from peft import MultitaskPromptTuningInit
+from peft import MultitaskPromptTuningInit, LoraConfig
 from transformers import AutoConfig
 
 # First Party
@@ -51,7 +51,7 @@ class TuningType(str, Enum):
     # MULTITASK_PREFIX_TUNING = "MULTITASK_PREFIX_TUNING"
     # P_TUNING = "P_TUNING"
     # PREFIX_TUNING = "PREFIX_TUNING"
-    # LORA = "LORA"
+    LORA = "LORA"
 
 
 def resolve_base_model(base_model, cls, torch_dtype):
@@ -79,7 +79,15 @@ def resolve_base_model(base_model, cls, torch_dtype):
 
 
 def get_peft_config(
-    tuning_type, tuning_config, base_model, cls, torch_dtype, verbalizer
+    tuning_type, tuning_config, base_model, cls, torch_dtype, verbalizer,
+        lora_r=8,
+        lora_target_modules=None,
+        lora_alpha=8,
+        lora_dropout=0.0,
+        lora_fan_in_fan_out=False,
+        lora_bias="none",
+        lora_modules_to_save=None
+
 ):
 
     if tuning_type not in TuningType._member_names_:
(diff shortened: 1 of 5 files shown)


ibm-peach-fish:
For the readme update: 

From @gkumbhat on slack 
> Me: could i get more specification on what readme.md you want updated, I'm having a hard time following the link on [169](https://github.com/caikit/caikit-nlp/issues/169)
> G: just this section: https://github.com/caikit/caikit-nlp/blob/main/README.md#:~:text=Salient%20Feature(s)-,Text%20Generation,-1.%20PeftPromptTuning
Particularly “Salient Feature(s)” part
> Me: The link in the issue looked like that too / I think you mean the text generation row of the table in the introduction section / and from the links i think you're pointing to the PeftPromptTuning, but all of that work was dismissed, the LoRA stuff lives in the text genertaion model / and hence i am lost with what you want me to put where
> G: oh yes / I was thinking it would be a mention after Fine-tuning Both modules above provide optimized inference capability using Text Generation Inference Server / that way its in point 2 for TextGeneration but we add a line stating this module can also do LoRA fine-tuning

Sometimes when I don't understand things, I'll just walk away for a bit then I 'get it' when I came back- often it works. This time it didn't. So I need more guidance on what to put where in the readme.md, but ott this PR is ready for review

chakrn:
cc: @Ssukriti 

Would be great to get your eyes on this

gkumbhat (changes requested):
Thanks for adding support for LoRA. Upon review a couple of things needs to be changed:
1. The input to train function needs to be a data model or built-in python data types. This is so that we can automatically then leverage those data types and type hints to generate a server spec servable via `caikit.runtime`. Under the hood, these parameter to `.train` function gets converted to a proto spec automatically and gets exposed to our gRPC and REST server.  Currently it it accepting `LoRAConfig`, which won't work with auto generation of server and thus won't get exposed in API.
2. The lora vector and lora based PEFT model gets generated after the model is already trained. So it looks like the LoRA vectors / weights are actually not getting modified / trained. 
3. Once we move the LoRAConfig parameters to data model type object,, we would also need to be filter and set "reasonable defaults" for those configuration parameters.
4. It looks like the lora config related changes in `peft_config.py` aren't actually getting used.
5. Upon looking at the implementation and considering generally how Lora is being referenced as part of "prompt tuning", we decided it would probably be better to move this to prompt tuning module and add a flag to save prompt vectors via merged weights or not. There will certainly be some things we would need to figure out to correctly hook it up for inferencing and its routing configuration. But that probably aligns more with how people are using it.

ibm-peach-fish:
Based on the requested changes, I think a complete refactor is in order- closing.

chakrn:
I would like to keep this open while Sukriti reviews

Ssukriti:
@gkumbhat since I do not have write access to caikit-nlp, could you create a branch in main repo and merge this PR to that branch ? I can then continue to commit to that branch too, instead of having to pull from another fork

Ssukriti:
added first PR to create data model and utilities https://github.com/caikit/caikit-nlp/pull/270 

Remaining work is being paused till we have more direction on priorities. closing PR , but will continue to refer to it for any further refactor

PR #213, opened 2023-09-28
Remove stop sequence default

tharapalanivel:
Don't use `eos_token` if user doesn't specify a `stop_sequence`, this is redundant for TGIS

### DIFF
--- a/caikit_nlp/toolkit/text_generation/tgis_utils.py
+++ b/caikit_nlp/toolkit/text_generation/tgis_utils.py
@@ -159,7 +159,6 @@ def validate_inf_params(
 
 def get_params(
     preserve_input_text,
-    eos_token,
     max_new_tokens,
     min_new_tokens,
     truncate_input_tokens,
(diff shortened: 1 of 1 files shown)


caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -177,8 +176,6 @@ def get_params(
     """Get generation parameters
 
     Args:
-        eos_token: str
```
gkumbhat, line 180: nit: looks like `preserve_input_text` is not declared in the docstrings here?

caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -205,7 +202,7 @@ def get_params(
         token_ranks=True,
     )
     stopping = generation_pb2.StoppingCriteria(
-        stop_sequences=stop_sequences or [eos_token] if eos_token else None,
+        stop_sequences=stop_sequences,
```
gkumbhat, line 205: if `stop_sequences` are empty / not provided by the user, does this still work with TGIS?

gkumbhat (approved):
LGTM

PR #216, opened 2023-10-02
proposal: use setuptools/setuptools_scm for building

dtrifiro:
This PR changes the build system from `flit` to `setuptools`+[`setuptools_scm`](https://github.com/pypa/setuptools_scm/). The main driver behind this PR is to simplify development by having `setuptools_scm` automatically set the module version at build/install time instead of having a fixed `0.0.1` version replaced by CI.

Another advantage of using `setuptools_scm` is to have proper versions reported when using editable installs, avoiding version requirements warnings/errors.

This also adds the possibility of using `caikit_nlp.__version__`  (or `__version_tuple__`) to check  the module version at runtime

### DIFF
--- a/.github/workflows/publish-library.yml
+++ b/.github/workflows/publish-library.yml
@@ -25,12 +25,11 @@ jobs:
       - uses: actions/checkout@v3
       - name: Set up Python
         uses: actions/setup-python@v3
-      - name: Release
-        env:
-          FLIT_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+      - name: Build and check package
         run: |
-          pip install tox
-          RELEASE_VERSION=${GITHUB_REF#refs/*/}
-          RELEASE_VERSION=${RELEASE_VERSION#v*}
-          sed -i "s/^version = .*/version = \"${RELEASE_VERSION}\"/" pyproject.toml
-          tox -e publish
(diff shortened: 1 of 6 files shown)


dtrifiro:
An example of what the version generated by `setuptools-scm`: `0.2.2.dev39+g8cfe9b8`:  it includes git tag, number of commits since last tag and the git ref for the build.

dtrifiro:
Hey @gabe-l-hart, do you have any bandwidth to have a look at this and/or discuss it?

gkumbhat (commented):
I like this change. @gabe-l-hart or @evaline-ju or @joerunde thoughts on this? Also do we want to include this for `caikit` as well?

Currently `caikit-nlp` cannot be published to pypi itself because of peft version, but something seems this would be useful for caikit and once we are able to push caikit-nlp to pypi.

joerunde (approved):
@dtrifiro yeah this looks better than what we whipped up when we pulled this all out into github. Thanks!

Can you go ahead and make the change to caikit as well?

joerunde:
Ah, I am not a code owner here so I'd actually need @gkumbhat to hit approve on this PR

gabe-l-hart (approved):
Looks good. Thank you!

PR #223, opened 2023-10-03
:goal_net: Disallow empty train streams

### LINKED ISSUE #214: Improve error handling on empty data streams
## Describe the bug

When running prompt tuning with an empty datastream, we get an obscure:
```
value check failed: Buffer size must be an int > 0
```
error back.

The [data stream wrapper](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/toolkit/data_stream_wrapper.py) sets the buffer length to the length of the data stream, but doesn't check up front if the stream is length 0, then fails later on. It would be a much more usable error message if we checked up-front and rejected empty streams.

evaline-ju:
- Add value checks on streams in `.train` - we were seeing confusing errors on empty streams in training. Specifically, an error check on buffer size will fail when the data stream wrapper is called/used
- Update test fixtures to not have empty streams

Closes #214

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
+++ b/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
@@ -348,6 +348,7 @@ class PeftPromptTuning(ModuleBase):
             PeftPromptTuning
                 Instance of this class with tuned prompt vectors.
         """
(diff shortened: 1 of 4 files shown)


gkumbhat (approved):
Thanks Evaline. LGTM!

PR #224, opened 2023-10-04
Add embedding task

markstur:
Adds an embedding-retrieval task to get embedding vector for a sentence.

- Return a list of floats.  For one string (a.k.a. sentence).
- Support for multiple strings removed for future batch support.
- Uses sentence-transformers.
- The data model allows for different float types (py, np.float32, np.float64).

Data model is based on Gabe's feedback in https://github.com/caikit/caikit-nlp/pull/39

This embedding service will be extended in separate PRs.  With multi-task support.  This service can also support sentence-similarity and reranking.

### DIFF
--- a/caikit_nlp/data_model/embedding_vectors.py
+++ b/caikit_nlp/data_model/embedding_vectors.py
@@ -16,8 +16,10 @@
 
 # Standard
 from typing import List, Union
+import json
 
 # Third Party
+from py_to_proto.dataclass_to_proto import Annotated, FieldNumber, OneofField
 import numpy as np
 
 # First Party
@@ -28,19 +30,35 @@ import alog
 log = alog.use_channel("DATAM")
 error = error_handler.get(log)
 
-# TODO:
-# At this point, py-to-proto==0.2.0 does not support other float types like float16 etc.
-# Once those types are supported in py-to-proto, we would want to enable those
-# as supported data types for Vector1D as well
+
+@dataobject(package="caikit_data_model.caikit_nlp")
+class FloatValue(DataObjectBase):
+    float_value: Union[
+        Annotated[float, OneofField("data_float"), FieldNumber(1)],
+        Annotated[np.float32, OneofField("data_float32"), FieldNumber(2)],
+        Annotated[np.float64, OneofField("data_float64"), FieldNumber(3)],
+    ]
 
 
 @dataobject(package="caikit_data_model.caikit_nlp")
 class Vector1D(DataObjectBase):
-    # NOTE: This currently only support float types
-    data: List[float]
+    """Data representation for a 1 dimension vector.
+    NOTE: This only support float types
+    """
+
+    data: List[FloatValue]
 
     def _convert_np_to_list(self, values):
         return values.tolist()
+
+    def _convert_to_floatvalue(self, values):
+
+        if isinstance(values, np.ndarray):
+            if values.dtype == np.float32:
+                return [FloatValue(data_float32=val).to_proto() for val in values]
+            if values.dtype == np.float64:
+                return [FloatValue(data_float64=val).to_proto() for val in values]
+        return [FloatValue(val).to_proto() for val in values]
 
     def to_dict(self) -> dict:
 
@@ -49,6 +67,49 @@ class Vector1D(DataObjectBase):
             return {"data": self._convert_np_to_list(self.data)}
 
         return {"data": self.data}
+
+    def fill_proto(self, proto):
+        data_proto = getattr(proto, "data")
+        data_proto.extend(self._convert_to_floatvalue(self.data))
+
+        return proto
+
+    @classmethod
+    def from_proto(cls, proto):
+        data = []
+        for datum in proto.data:
+            value_type = FloatValue.from_proto(datum).which_oneof("float_value")
+            data.append(getattr(datum, value_type))
+        return cls(**{"data": data})
+
+    @classmethod
+    def from_json(cls, json_str):
+        """Function to convert json_str to DataBase object
+
+        NOTE: Since JSON doesn't support variety of float formats
+              we would only be able to convert all the values to python floats here.
+              So there would be a loss of information when converting to and from JSON
+
+        Args:
+            json_str: str or dict
+                A stringified JSON specification/dict of the data_model
+
+        Returns:
+            caikit.core.data_model.DataBase
+                A DataBase object.
+        """
+        error.type_check("<NLP97633962E>", dict, str, json_str=json_str)
+        if isinstance(json_str, str):
+            json_str = json.loads(json_str)
+
+        error.value_check(
+            "<NLP49158301E>", "data" in json_str, "invalid json_str provided!"
+        )
+
+        data = []
+        for datum in json_str["data"]:
+            data.append(datum)
+        return cls(data=data)
 
 
 @dataobject(package="caikit_data_model.caikit_nlp")
@@ -112,3 +173,16 @@ class EmbeddingResult(DataObjectBase):
         vector_proto.extend([val.to_proto() for val in proto_values])
 
         return proto
+
+    @classmethod
+    def from_json(cls, json_str):
+        error.type_check("<NLP83742499E>", dict, str, json_str=json_str)
+
+        if isinstance(json_str, str):
+            json_str = json.loads(json_str)
+
+        error.value_check(
+            "<NLP64229346E>", "data" in json_str, "invalid json_str provided!"
+        )
+
+        return cls(data=[Vector1D.from_json(datum) for datum in json_str["data"]])
--- a/caikit_nlp/data_model/embedding_vectors.py
+++ b/caikit_nlp/data_model/embedding_vectors.py
@@ -16,15 +16,13 @@
 
 # Standard
 from typing import List, Union
-import json
 
 # Third Party
-from py_to_proto.dataclass_to_proto import Annotated, FieldNumber, OneofField
 import numpy as np
 
 # First Party
 from caikit.core import DataObjectBase, dataobject
-from caikit.core.toolkit import error_handler
+from caikit.core.exceptions import error_handler
 import alog
 
 log = alog.use_channel("DATAM")
@@ -32,84 +30,88 @@ error = error_handler.get(log)
 
 
 @dataobject(package="caikit_data_model.caikit_nlp")
-class FloatValue(DataObjectBase):
-    float_value: Union[
-        Annotated[float, OneofField("data_float"), FieldNumber(1)],
-        Annotated[np.float32, OneofField("data_float32"), FieldNumber(2)],
-        Annotated[np.float64, OneofField("data_float64"), FieldNumber(3)],
-    ]
+class PyFloatSequence(DataObjectBase):
+    values: List[float]
+
+
+@dataobject(package="caikit_data_model.caikit_nlp")
+class NpFloat32Sequence(DataObjectBase):
+    values: List[np.float32]
+
+
+@dataobject(package="caikit_data_model.caikit_nlp")
+class NpFloat64Sequence(DataObjectBase):
+    values: List[np.float64]
 
 
 @dataobject(package="caikit_data_model.caikit_nlp")
 class Vector1D(DataObjectBase):
-    """Data representation for a 1 dimension vector.
-    NOTE: This only support float types
-    """
+    """Data representation for a 1 dimension vector of float-type data."""
 
-    data: List[FloatValue]
+    data: Union[
+        PyFloatSequence,
+        NpFloat32Sequence,
+        NpFloat64Sequence,
+    ]
 
-    def _convert_np_to_list(self, values):
-        return values.tolist()
-
-    def _convert_to_floatvalue(self, values):
-
-        if isinstance(values, np.ndarray):
-            if values.dtype == np.float32:
-                return [FloatValue(data_float32=val).to_proto() for val in values]
-            if values.dtype == np.float64:
-                return [FloatValue(data_float64=val).to_proto() for val in values]
-        return [FloatValue(val).to_proto() for val in values]
+    # Classes to wrap values for one-of
+    WRAP_MAP = {
+        "data_pyfloatsequence": PyFloatSequence,
+        "data_npfloat32sequence": NpFloat32Sequence,
+        "data_npfloat64sequence": NpFloat64Sequence,
+    }
+    WRAP_DEFAULT = PyFloatSequence
 
     def to_dict(self) -> dict:
+        woo = self.which_oneof("data")  # determine which one of to set
+        if not woo:
+            woo = "data"
+            values = self.data
+        else:
+            values = self.data.values
 
-        # If data is in np.ndarray format, convert it to python list
-        if isinstance(self.data, np.ndarray):
-            return {"data": self._convert_np_to_list(self.data)}
-
-        return {"data": self.data}
-
-    def fill_proto(self, proto):
-        data_proto = getattr(proto, "data")
-        data_proto.extend(self._convert_to_floatvalue(self.data))
-
-        return proto
+        return {
+            woo: {
+                # coerce numpy.ndarray and numpy.float32 into JSON serializable list of floats
+                "values": [float(x) for x in values]
+            }
+        }
 
     @classmethod
     def from_proto(cls, proto):
-        data = []
-        for datum in proto.data:
-            value_type = FloatValue.from_proto(datum).which_oneof("float_value")
-            data.append(getattr(datum, value_type))
-        return cls(**{"data": data})
-
-    @classmethod
-    def from_json(cls, json_str):
-        """Function to convert json_str to DataBase object
-
-        NOTE: Since JSON doesn't support variety of float formats
-              we would only be able to convert all the values to python floats here.
-              So there would be a loss of information when converting to and from JSON
-
-        Args:
-            json_str: str or dict
-                A stringified JSON specification/dict of the data_model
-
-        Returns:
-            caikit.core.data_model.DataBase
-                A DataBase object.
-        """
-        error.type_check("<NLP97633962E>", dict, str, json_str=json_str)
-        if isinstance(json_str, str):
-            json_str = json.loads(json_str)
-
-        error.value_check(
-            "<NLP49158301E>", "data" in json_str, "invalid json_str provided!"
+        woo_key = "data"
+        woo = proto.WhichOneof(woo_key)
+        woo_data = getattr(proto, woo)
+        return cls(
+            **{woo_key: (cls.WRAP_MAP.get(woo, cls.WRAP_DEFAULT)(woo_data.values))}
         )
 
-        data = []
-        for datum in json_str["data"]:
-            data.append(datum)
-        return cls(data=data)
+    def fill_proto(self, proto):
+
+        if hasattr(self.data, "values"):
+            values = self.data.values
+        else:
+            values = self.data
+
+        if len(values) > 0:
+            sample = values[0]
+            if isinstance(sample, np.float64):
+                proto.data_npfloat64sequence.values.extend(
+                    list(np.array(values, dtype=np.float64))
+                )
+            elif isinstance(sample, np.float32):
+                proto.data_npfloat32sequence.values.extend(
+                    list(np.array(values, dtype=np.float32))
+                )
+            elif isinstance(sample, float):
+                proto.data_pyfloatsequence.values.extend(values)
+            else:
+                error(
+                    "<NLP58452707E>",
+                    "unsupported type of data value. Expecting a float type",
+                )
+
+        return proto
 
 
 @dataobject(package="caikit_data_model.caikit_nlp")
@@ -117,72 +119,3 @@ class EmbeddingResult(DataObjectBase):
     """Data representation for an embedding matrix holding 2D vectors"""
 
     data: List[Vector1D]
-
-    def _convert_np_to_vector1d(self, values):
-        if values.ndim == 1:
-            return [Vector1D(values)]
-
-        if values.ndim == 2:
-            return [Vector1D(vector) for vector in values]
-
-        error(
-            "<NLP18185885E>",
-            f"Unsupported dimensions {values.ndim} for EmbeddingResult",
-        )
-
-    def _convert_list_to_vector1d(self, values):
-        temp_data = None
-        if isinstance(values[0], list):
-            temp_data = [Vector1D(vector) for vector in values]
-        else:
-            temp_data = [Vector1D(values)]
-
-        return temp_data
-
-    def to_dict(self) -> dict:
-
-        vector_2d: Union(List[Vector1D], None) = None
-
-        # if data is in np.ndarray format, convert it into Vector1D
-        if isinstance(self.data, np.ndarray):
-            vector_2d = self._convert_np_to_vector1d(self.data)
-        elif isinstance(self.data, list):
-            vector_2d = self._convert_list_to_vector1d(self.data)
-        else:
-            error(
-                "<NLP69094934E>",
-                f"Only list and np.ndarray supported currently. Provided {type(self.data)}",
-            )
-
-        return {"data": [vector.to_dict() for vector in vector_2d]}
-
-    def fill_proto(self, proto):
-        vector_proto = getattr(proto, "data")
-
-        proto_values = None
-        if isinstance(self.data, list):
-            proto_values = self._convert_list_to_vector1d(self.data)
-        elif isinstance(self.data, np.ndarray):
-            proto_values = self._convert_np_to_vector1d(self.data)
-        else:
-            error(
-                "<NLP58452707E>",
-                "unsupported type for data. Only list and np.ndarray supported",
-            )
-
-        vector_proto.extend([val.to_proto() for val in proto_values])
-
-        return proto
-
-    @classmethod
-    def from_json(cls, json_str):
-        error.type_check("<NLP83742499E>", dict, str, json_str=json_str)
-
-        if isinstance(json_str, str):
-            json_str = json.loads(json_str)
-
-        error.value_check(
-            "<NLP64229346E>", "data" in json_str, "invalid json_str provided!"
-        )
-
-        return cls(data=[Vector1D.from_json(datum) for datum in json_str["data"]])
--- a/caikit_nlp/data_model/embedding_vectors.py
+++ b/caikit_nlp/data_model/embedding_vectors.py
@@ -78,13 +78,13 @@ class Vector1D(DataObjectBase):
         )
 
     @classmethod
-    def from_embeddings(cls, embeddings):
-        if embeddings.dtype == np.float32:
-            data = NpFloat32Sequence(embeddings)
-        elif embeddings.dtype == np.float64:
-            data = NpFloat64Sequence(embeddings)
+    def from_vector(cls, vector):
+        if vector.dtype == np.float32:
+            data = NpFloat32Sequence(vector)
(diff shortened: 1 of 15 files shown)


markstur:
Tag @gabe-l-hart and will tag Anjali if I can find her.

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,140 @@
...
+            EmbeddingModule
+                Instance of this class built from the model.
+        """
+
+        config = ModuleConfig.load(model_path)
+        load_path = config.get("model_artifacts")
+
+        artifact_path = False
+        if load_path:
+            if os.path.isabs(load_path) and os.path.isdir(load_path):
```
gkumbhat, line 71: any specific reason to allow absolute path as a value inside the `config.yml` ? Usually this is a relative path to the `model_path`

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,140 @@
...
+                artifact_path = load_path
+            else:
+                full_path = os.path.join(model_path, load_path)
+                if os.path.isdir(full_path):
+                    artifact_path = full_path
+
+        if not artifact_path:
```
gkumbhat, line 78: hmm, since `artifact_path` always depends on `load_path` which would be None if `model_artifacts` doesn't exist, can we remove `artifact_path` check and instead use `load_path` directly ?

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,140 @@
...
+            artifact_path = config.get("hf_model", DEFAULT_HF_MODEL)
```
gkumbhat, line 79: A little confused, how is `ht_model` value actually getting into the `config` ? This isn't getting set in `.save` method below?

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,140 @@
...
+
+# Local
+from .embedding_retrieval_task import EmbeddingRetrievalTask
+from caikit_nlp.data_model.embedding_vectors import EmbeddingResult, Vector1D
+
+logger = alog.use_channel("<EMBD_BLK>")
+error = error_handler.get(logger)
+
+HOME = Path.home()
+DEFAULT_HF_MODEL = "sentence-transformers/all-MiniLM-L6-v2"
```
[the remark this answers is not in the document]
gkumbhat: Modules usually doesn't work with default model paradigm, what is the particular use-case for this?

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,140 @@
...
+        saver = ModuleSaver(
+            self,
+            model_path=model_path,
+        )
+
+        # Extract object to be saved
+        with saver:
+            saver.update_config({"model_artifacts": "."})
+            if self.model:  # This condition allows for empty placeholders
+                self.model.save(model_path)
```
gkumbhat, line 140: We would always want to save the model right? Whats the use-case to generate empty config object instead of model?

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,140 @@
...
+                if os.path.isdir(full_path):
+                    artifact_path = full_path
+
+        if not artifact_path:
+            artifact_path = config.get("hf_model", DEFAULT_HF_MODEL)
+
+        return cls.bootstrap(artifact_path)
+
+    def run(
+        self, input: List[str], **kwargs  # pylint: disable=redefined-builtin
```
gkumbhat, line 84: Usual paradigm is for `.run` function to accept `str` and `run_batch` function to accept `List[str]` can we split this into 2 functions ?

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "pandas>=1.5.0",
     "scikit-learn>=1.1",
     "scipy>=1.8.1",
+    "sentence-transformers~=2.2.2",
```
gkumbhat, line 27: Not that we can do anything about it, but `sentence-transformers` also brings in `torchvision` as dependency, which is unnecessary for NLP / Text use-cases. 😢

gkumbhat (changes requested):

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,140 @@
...
+    def bootstrap(cls, base_model_path: str) -> "EmbeddingModule":
+        """Bootstrap a HuggingFace transformer-based sequence classification model
+
+        Args:
+            base_model_path: str
+                Path to the model to be loaded.
+        """
+        model = SentenceTransformer(
+            base_model_path,
+            cache_folder=f"{HOME}/.cache/huggingface/sentence_transformers",
```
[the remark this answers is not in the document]
markstur: yep.  We were just talking about this one the other day.  Definitely should use that env.  I'll just need to do some testing with the default.

caikit_nlp/modules/embedding_retrieval/embedding.py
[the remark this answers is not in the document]
markstur: This is a convenience (think demo), but I think that doesn't apply to caikit-nlp.  I'll make this fail if the config doesn't have model.

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: Yeah, the else would handle the relative path which might be more common, but since we're looking at a separate volume with just models it would be good to not require ../../ to go up above the models (configs) folder.

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: right. I think for readability I should just use it and overwrite it.

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: I'll have to fix save.

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: Just fyi this is because we always edit the configs and don't use save.

The save/load test passed but I suppose the default model allowed it to pass when it should not have.

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: We'll have to discuss this.  It would be nice to have a str | List[str], but also the ability to send one gRPC/REST request with a bunch of inputs.

pyproject.toml
markstur: bummer.  I didn't notice that.  I thought we did good to avoid some other baggage.  :(


https://github.com/UKPLab/sentence-transformers/pull/1881

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: It turns out that join() already handles the absolute path checking (ignores parts up to the abspath).

caikit_nlp/modules/embedding_retrieval/embedding.py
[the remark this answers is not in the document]
rzbhatti: Ideally, we would like to make the specification of list of models (hugging face or local) part of the deployment time parameters rather than hard-coded in the service.

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: I refactored to clarify, but ended up keeping a 2nd variable to help clarify that we can use artifact_path or hf_model (next comment is about hf_model).

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: So I did a bunch of fixin of save(), but re: hf_model...  How it gets in there is someone edits the config.yml which is how we usually do it because there isn't really a great mechanism that I see anywhere for bootstrap+save for non-coders [that's a different issue].   Where hf_model is particularly useful is BYO model download for HF. This is pretty common for dev and allows the cache to be used.  BUT related to this comment I did some work on save() to make it safer but also it will save the model and set artifact_path, so hf_model is kind of a bootstrap or dev mechanism.

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: I think this is good as-is until we get multi-transaction way of doing both here.

caikit_nlp/modules/embedding_retrieval/embedding.py
markstur: Use case was to generate a config.yml for editing (e.g. add hf_model: model_name and run), but even without that this check avoids need for error check for self.model is None here when that responsibility is in other parts of the code.

caikit_nlp/modules/embedding_retrieval/embedding.py
gkumbhat: Created follow-up issue [here](https://github.com/caikit/caikit/issues/527)

caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,176 @@
...
+    values: List[np.float32]
+
+    @classmethod
+    def from_proto(cls, proto):
+        values = [np.float32(x) for x in proto.values]
+        return cls(values)
+
+
+@dataobject(package="caikit_data_model.caikit_nlp")
+@dataclass
```
gkumbhat, line 52: nit: we can likely remove this. `float64` seems too big for our use-case. Is this getting used by any embedding models?

caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+    @classmethod
+    def from_json(cls, json_str):
+        """JSON does not have different float types. Move data into data_pyfloatsequence"""
+
+        json_obj = json.loads(json_str) if isinstance(json_str, str) else json_str
+        data = json_obj.pop("data")
+        if data is not None:
+            json_obj["data_pyfloatsequence"] = data
```
gkumbhat, line 87: this seems a bit hacky. Looks like it would only assign value to `PyFloatSequence` as well right? 

@prashantgupta24 any recommendation?

markstur: Yes this is very unfortunate.  I don't see to/from json working with the vector union without this, but 🤮

caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,176 @@
...
+        except json_format.ParseError as ex:
+            error("<NLP39795399E>", ValueError(ex))
+
+    def to_dict(self) -> dict:
+        """to_dict is needed to make things serializable"""
+        values = self.data.values if self.data.values is not None else []
+        return {
+            "data": {
+                # coerce numpy.ndarray and numpy.float32 into JSON serializable list of floats
+                "values": [float(x) for x in values]
```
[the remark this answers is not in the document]
markstur: This will cause a JSON serializable error if a **List** of np.**float32** gets through, but that does not seem to be a problem with any normal use path so I'll use your optimization.

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,175 @@
...
+
+        artifacts_path = config.get(cls._ARTIFACTS_PATH_KEY)
+        if artifacts_path:
+            model_name_or_path = os.path.abspath(
+                os.path.join(model_path, artifacts_path)
+            )
+            error.dir_check("<NLP34197772E>", model_name_or_path)
+        else:
+            # If no artifacts_path, look for hf_model Hugging Face model by name (or path)
+            model_name_or_path = config.get(cls._HF_HUB_KEY)
```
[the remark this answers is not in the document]
markstur: it did, but I am removing all the hf_model download stuff per other review comments.  We should bring it back in some supported way in a separate PR I think 🤷‍♂️

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,175 @@
...
+        with saver:
+            artifacts_path = saver.config.get(self._ARTIFACTS_PATH_KEY)
+            if not artifacts_path:
+                artifacts_path = self._ARTIFACTS_PATH_DEFAULT
+                saver.update_config({self._ARTIFACTS_PATH_KEY: artifacts_path})
+            if self.model:  # This condition allows for empty placeholders
+                artifacts_path = os.path.abspath(
+                    os.path.join(model_path, artifacts_path)
+                )
+                self.model.save(artifacts_path, create_model_card=True)
```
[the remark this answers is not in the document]
markstur: ---
pipeline_tag: sentence-similarity
tags:
- sentence-transformers
- feature-extraction
- sentence-similarity
- transformers

---

# {MODEL_NAME}

This is a [sentence-transformers](https://www.SBERT.net) model: It maps sentences & paragraphs to a 32 dimensional dense vector space and can be used for tasks like clustering or semantic search.



## Usage (Sentence-Transformers)

Using this model becomes easy when you have [sentence-transformers](https://www.SBERT.net) installed:

```

File: caikit_nlp/modules/embedding_retrieval/embedding.py (hunk 3 above)
Comment (markstur) [reply caikit_nlp/modules/embedding_retrieval/embedding.py]: hf_model stuff removed

File: caikit_nlp/modules/embedding_retrieval/embedding.py (hunk 6 above)
Comment (markstur) [reply caikit_nlp/modules/embedding_retrieval/embedding.py]: latest version is single str only.  Will work on multi-task (the multi-"transaction" mis-speak is my bad) in separate PR.

File: caikit_nlp/data_model/embedding_vectors.py (hunk 10 above)
Comment (markstur) [reply caikit_nlp/data_model/embedding_vectors.py:97]: also note:  it is only to pyfloat because coming from JSON it's just a number.  Not sure who would want a data object from JSON, but testing the to/from implementation requires a solution here.

File: caikit_nlp/modules/embedding_retrieval/embedding.py [hunk 14]
```diff
@@ -0,0 +1,175 @@
...
+        result = self.model.encode(sentences)
+
+        vectors: List[Vector1D] = []
+        for vector in result:
+            if vector.dtype == np.float32:
+                vectors.append(Vector1D(NpFloat32Sequence(vector)))
+            elif vector.dtype == np.float64:
+                vectors.append(Vector1D(NpFloat64Sequence(vector)))
+            else:
+                vectors.append(Vector1D(PyFloatSequence(vector)))
```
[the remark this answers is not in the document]
markstur: done.  Looks pretty good that way

caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @classmethod
+    def from_proto(cls, proto):
+        values = [np.float32(x) for x in proto.values]
+        return cls(values)
+
+
+@dataobject(package="caikit_data_model.caikit_nlp")
+@dataclass
+class NpFloat64Sequence(DataObjectBase):
```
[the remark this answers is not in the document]
markstur: I think we agreed "might as well keep this" so the data model can be used for future addition of fp16.  The current module doesn't need this, but enhancements or future modules using the same interface would.  Otherwise List[float] is much easier.

markstur:
rebased to catch up to main (and so I can stack PRs with multi-task support)

caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @classmethod
+    def from_proto(cls, proto):
+        values = np.asarray(proto.values, dtype=np.float64)
+        return cls(values)
+
+
+@dataobject(package="caikit_data_model.caikit_nlp")
+@dataclass
+class Vector1D(DataObjectBase):
```
gkumbhat, line 64: nit: we usually try to name the output object name conveying the "task" related output. Since this is directly output of EmbeddingTask, can we rename this to `EmbeddingVector` ?

caikit_nlp/data_model/embedding_vectors.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def __post_init__(self):
+        error.value_check(
+            "<NLP92989048E>",
+            hasattr(self.data, "values"),
+            ValueError("Vector1D requires a float sequence data object with values."),
+        )
+
+    @classmethod
+    def from_embeddings(cls, embeddings):
```
gkumbhat, line 81: nit: Can we rename this function to `from_vector`.

evaline-ju (commented):
few small things

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,143 @@
...
+        model_config_path = model_path  # because the param name is misleading
+
+        error.type_check("<NLP82314992E>", str, model_path=model_config_path)
+        error.value_check(
+            "<NLP40145207E>",
+            model_config_path is not None and model_config_path.strip(),
+            f"model_path '{model_config_path}' is invalid",
+        )
+
+        model_config_path = os.path.abspath(
```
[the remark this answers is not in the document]
markstur: The ModuleSaver docstring says it takes an "absolute path":
``` 
model_path (str): The absolute path to the directory where the model
           will be saved.  If this directory does not exist, it will be
           created.
```
It doesn't look like that is necessary today, but better to follow the doc.

tests/modules/embedding_retrieval/test_embedding_retrieval.py
```diff
@@ -0,0 +1,96 @@
+"""Tests for sequence classification module
```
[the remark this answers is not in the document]
markstur: Fixed comment.  Renamed test file to match.  Did some other renaming for more consistency.

tests/data_model/test_embedding_vectors.py
```diff
@@ -0,0 +1,120 @@
...
+
+def test_vector1d_iterator_error():
+    """Cannot just shove in an iterator and expect it to work"""
+    with pytest.raises(ValueError):
+        dm.Vector1D(data=[1.1, 2.2, 3.3])
+
+
+def test_vector1d_trick():
+    """FYI -- The param check currently allows for objects with values using this trick"""
+    dm.Vector1D(data=TRICK_SEQUENCE(values=[1.1, 2.2]))
```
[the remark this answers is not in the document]
markstur: I removed this test.  It was redundant with below where the results are validated.  Less confusing to just get rid of this one.

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,143 @@
...
+        )
+
+        # Get and update config (artifacts_path)
+        artifacts_path = saver.config.get(self._ARTIFACTS_PATH_KEY)
+        if not artifacts_path:
+            artifacts_path = self._ARTIFACTS_PATH_DEFAULT
+            saver.update_config({self._ARTIFACTS_PATH_KEY: artifacts_path})
+
+        # Save the model
+        artifacts_path = os.path.abspath(
```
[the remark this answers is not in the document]
markstur: Nope.  Removed.

caikit_nlp/modules/embedding_retrieval/embedding.py
```diff
@@ -0,0 +1,143 @@
...
+# First Party
+from caikit.core import ModuleBase, ModuleConfig, ModuleSaver, module
+from caikit.core.exceptions import error_handler
+import alog
+
+# Local
+from .embedding_retrieval_task import EmbeddingRetrievalTask
+from caikit_nlp.data_model.embedding_vectors import Vector1D
+
+logger = alog.use_channel("<EMBD_BLK>")
```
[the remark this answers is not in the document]
markstur: Thanks!

caikit_nlp/data_model/embedding_vectors.py
markstur: Done

caikit_nlp/data_model/embedding_vectors.py
markstur: Changed to EmbeddingResponse with Vector1D in it.

My opinion?  I tend to dislike these extra levels like `resp.result.data.values[2]` when resp[2] would've been much nicer considering we really just want a List[float] (more or less).  But I guess it looks pretty good this new way.

caikit_nlp/modules/embedding_retrieval/embedding.py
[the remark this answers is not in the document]
markstur: re: example using ModuleSaver...  I'd recommend you do not use the ModuleSaver context manager.  It isn't safe. https://github.com/caikit/caikit/issues/525

markstur:
@gkumbhat anything I need to do to unblock this one?   Not sure if it helps if click the "resolve conversation" buttons.  Up to you.

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -0,0 +1,142 @@
...
+# Local
+from .embedding_tasks import EmbeddingTask
+from caikit_nlp.data_model.embedding_vectors import EmbeddingResult, Vector1D
+
+logger = alog.use_channel("TXT_EMB")
+error = error_handler.get(logger)
+
+
+@module(
+    "EEB12558-B4FA-4F34-A9FD-3F5890E9CD3F",
```
gkumbhat, line 35: interesting change I noticed that the id here contains all caps (plus numbers).. Other modules have small cases. There isn't a rule enforcing around it, but might be good to be consistent. 🤔

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -0,0 +1,142 @@
...
+        """Run inference on model.
+        Args:
+            input: str
+                Input text to be processed
+        Returns:
+            EmbeddingResult: the result vector nicely wrapped up
+        """
+        error.type_check("<NLP27491611E>", str, input=input)
+
+        return EmbeddingResult(Vector1D.from_vector(self.model.encode(input)))
```
gkumbhat, line 91: do we need to do some validation on `model.encode` output before converting it to embedding via vector or is it guaranteed to be a compatible vector always?

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -0,0 +1,142 @@
...
+        )  # No leading/trailing spaces sneaky weirdness
+
+        os.makedirs(model_config_path, exist_ok=False)
+        saver = ModuleSaver(
+            module=self,
+            model_path=model_config_path,
+        )
+
+        # Get and update config (artifacts_path)
+        artifacts_path = saver.config.get(self._ARTIFACTS_PATH_KEY)
```
gkumbhat, line 131: this will always be none right? Unless we are overriding the model ?

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -0,0 +1,142 @@
...
+        if not artifacts_path:
+            artifacts_path = self._ARTIFACTS_PATH_DEFAULT
+            saver.update_config({self._ARTIFACTS_PATH_KEY: artifacts_path})
+
+        # Save the model
+        self.model.save(
+            os.path.join(model_config_path, artifacts_path), create_model_card=True
```
gkumbhat, line 138: how does the model card gets saved  and consumed?

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -0,0 +1,142 @@
...
+            "<NLP40145207E>",
+            model_config_path is not None and model_config_path.strip(),
+            f"model_path '{model_config_path}' is invalid",
+        )
+
+        model_config_path = os.path.abspath(
+            model_config_path.strip()
+        )  # No leading/trailing spaces sneaky weirdness
+
+        os.makedirs(model_config_path, exist_ok=False)
```
gkumbhat, line 124: We can also use `saver.add_dirs` functions from here: https://github.com/caikit/caikit/blob/a16f063f6155f0088eb9959a32b7f0871e89731d/caikit/core/modules/saver.py#L117

This will avoid need to also make the absolute path etc above.

gkumbhat: We can also use the context base use of `ModuleSaver` and that can take care of adding / making the directory as well.  Example: https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L472

gkumbhat (commented):
Left some small things, but other than that it looks good. Thanks @markstur

caikit_nlp/modules/text_embedding/embedding.py
markstur: > We can also use `saver.add_dirs` functions from here: https://github.com/caikit/caikit/blob/a16f063f6155f0088eb9959a32b7f0871e89731d/caikit/core/modules/saver.py#L117
> 
> This will avoid need to also make the absolute path etc above.

os.makedirs() is used because in-place updates are not safe to run on existing dirs.  I use exist_ok=False to enforce that where ModuleSaver would not and can remove existing file trees on exceptions.  Not cool.

The abspath() is per ModuleSaver param docstrings.  strip() should not be necessary but seems better than testing what blanks would do for user experience.

markstur: > We can also use the context base use of `ModuleSaver` and that can take care of adding / making the directory as well. Example: https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/peft_prompt_tuning.py#L472

context for ModuleSaver is pretty dangerous here.  Use save() wrong and ModuleSaver will wipe out a directory that might be important.

caikit_nlp/modules/text_embedding/embedding.py
markstur: right, I can remove this

caikit_nlp/modules/text_embedding/embedding.py
markstur: I'll remove the option, but it'a a model card that is typically stored with the model files documenting what it there.

caikit_nlp/modules/text_embedding/embedding.py
markstur: Always returns a tensor that works with Vector1D.from_vector().   There is nothing to validate here.  If the dtype is something we didn't expect we use PyFloatSequence().   I could add a __post_init__ so PyFloatSequence() evaluates each value, but I don't have a real case for that right now.

caikit_nlp/modules/text_embedding/embedding.py
markstur: done (in-coming)

markstur:
@gkumbhat Updated based on feedback.  I don't want to use some of the saver() code until it is fixed to be safe and even revert partial changes, but we did get some unnecessary stuff out of save().  Thanks!

caikit_nlp/modules/text_embedding/embedding.py
gkumbhat: but that is related to `os.mkdirs` only right? So that can be fixed by adding this an option in the context manager or in the caikit repo itself? May be club it with your DM PR in caikit ?

gkumbhat (approved):
LGTM

PR #228, opened 2023-10-07
model_run_utils: return correct input_token_count

dtrifiro:
`input_token_count` currently is always `2`  because `tok_tensors` is a `dict`


gkumbhat (approved):
LGTM. Thanks for catching this

PR #231, opened 2023-10-11
Make verbalizer rendering optional in eval script

alex-jw-brooks:
Currently, the evaluation script breaks on non-prompt tuned models, because the local text generation module doesn't have a verbalizer property. This change enables the script to work on local text generation models by making the verbalized text writing conditional on whether or not a verbalizer attribute exists.

### DIFF
--- a/examples/evaluate_model.py
+++ b/examples/evaluate_model.py
@@ -120,15 +120,15 @@ def export_model_preds(preds_file, predictions, validation_stream, verbalizer):
     """
     pred_objs = []
     for pred, record in zip(predictions, validation_stream):
-        src, target = record.input, record.output
-        pred_objs.append(
-            {
-                "source": record.input,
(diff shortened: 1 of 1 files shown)


examples/evaluate_model.py
```diff
@@ -160,7 +160,10 @@ def export_model_preds(preds_file, predictions, validation_stream, verbalizer):
         )
     )
     export_model_preds(
-        args.preds_file, predictions, validation_stream, model.verbalizer
+        args.preds_file,
+        predictions,
+        validation_stream,
+        getattr(model, "verbalizer", None),
```
gkumbhat, line 166: QQ: model will not contain `verbalizer` in case of  non-prompt tuning ones, right? (Just checking)

gkumbhat (approved):
LGTM. Left a quick clarification question

PR #234, opened 2023-10-12
Revert ":goal_net: Disallow empty train streams"

gkumbhat:
Reverts caikit/caikit-nlp#223

Context: some attributes like length may not be present when attempting to create `DataStream` from `DataStreamSource`. Pending further investigation on the `caikit` side for this - we are reverting this PR to unblock

### DIFF
--- a/tests/modules/text_generation/test_peft_prompt_tuning.py
+++ b/tests/modules/text_generation/test_peft_prompt_tuning.py
@@ -216,7 +216,7 @@ def test_train_model_classification_record(causal_lm_train_kwargs, set_cpu_devic
 
 
 def test_prompt_output_types(causal_lm_train_kwargs):
-    # Try training a model with outpout_model_types set to a list of strings
+    # Try training a model with output_model_types set to a list of strings
     patch_kwargs = {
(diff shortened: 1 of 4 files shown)


tests/modules/text_generation/test_peft_prompt_tuning.py
```diff
@@ -216,7 +216,7 @@ def test_train_model_classification_record(causal_lm_train_kwargs, set_cpu_devic
 
 
 def test_prompt_output_types(causal_lm_train_kwargs):
-    # Try training a model with output_model_types set to a list of strings
+    # Try training a model with outpout_model_types set to a list of strings
```
evaline-ju, line 219: heh unfortunate re-typo

gkumbhat, line 219:
```suggestion
    # Try training a model with output_model_types set to a list of strings
```

evaline-ju (approved):
LGTM - we should [eventually] have checks for empty streams but worth more investigation time from the `caikit` side as noted in the PR description

PR #236, opened 2023-10-17
Update to tasks

tharapalanivel:
To enable machine learning models that support multiple tasks, `caikit` has updated the `mdl.TASK_CLASS` parameter to `type(mdl).tasks` which is now a set of all tasks supported by that particular model. This PR upgrades `caikit` and subsequently fixes the `FilteredSpanClassification` module's task-based validation.

### DIFF
--- a/caikit_nlp/modules/token_classification/filtered_span_classification.py
+++ b/caikit_nlp/modules/token_classification/filtered_span_classification.py
@@ -106,6 +106,11 @@ class FilteredSpanClassification(ModuleBase):
             any(tasks_intersection),
             f"classifier does not implement one of required tasks: {ALLOWED_TASKS}",
         )
+        error.value_check(
+            "<NLP41319815E>",
+            len(tasks_intersection) == 1,
+            f"classifier should implement only one task in: {ALLOWED_TASKS}",
+        )
         self.lang = lang
         self.tokenizer = tokenizer
         self.classifier = classifier


caikit_nlp/modules/token_classification/filtered_span_classification.py
```diff
@@ -87,7 +87,7 @@ def __init__(
         error.type_check("<NLP79642537E>", ModuleBase, tokenizer=tokenizer)
         error.value_check(
             "<NLP42736791E>",
-            tokenizer.TASK_CLASS == TokenizationTask,
+            TokenizationTask in tokenizer._TASK_CLASSES,
```
evaline-ju, line 90: could we check `tasks`  now as opposed to checking the private attribute? ref. https://github.com/caikit/caikit/blob/main/caikit/core/modules/decorator.py#L157

caikit_nlp/modules/token_classification/filtered_span_classification.py
```diff
@@ -99,18 +99,24 @@ def __init__(
...
         error.value_check(
             "<NLP41319814E>",
-            classification_task in ALLOWED_TASKS,
+            any([i for i in classification_tasks if i in ALLOWED_TASKS]),
             f"classifier does not implement one of required tasks: {ALLOWED_TASKS}",
         )
+        classification_task_set = set(classification_tasks)
+        error.value_check(
+            "<NLP41319815E>",
+            len(classification_task_set) == 1,
```
evaline-ju, line 111: From my understanding the goal here is to check if the classifier implements one but not more than one of the ALLOWED_TASKS - can we just check that both (in this case) are not in the ALLOWED_TASKS?

I think this would disallow a case where a multitask classifier may implement, say TextClassification and a theoretical say, SentimentClassification task (that would have no effect on the particular module here)? (might be good to have a small test addition)

evaline-ju: I'm a bit confused - by "unnecessary" did you mean to remove it?

I'm thinking of the case where [attributes replaced with strings for simplicity]
```
classification_tasks = ['sentiment', 'textclassification'] # from type(classifier).tasks but for simplicity
allowed_tasks = ['textclassification', 'tokenclassification']
```
the `any([i for i in classification_tasks....` should return True as expected but the
```
len(set(classification_tasks)) == 1
```
will return False for the classifier. If the "sentiment" task is perhaps just a more specific implementation, for example, for the text classification task, it probably should not trigger an error and be disallowed from use in this module?

evaline-ju (approved):
LGTM!

PR #239, opened 2023-10-18
pyproject: use setuptools' automatic discovery

dtrifiro:
By not explicitly defining packages, setuptools will automatically find all packages using the flat-layout strategy.
    
This also prevents the following warnings to be printed when building using `python -m build`:
```
 _Warning: Package 'caikit_nlp.toolkit' is absent from the `packages` configuration.
```


dtrifiro:
Note that the linting error is (somehow) not related to this change. `__init__.py` was missing in the text_generation submodule and thus  pylint was not running in that directory.

PR #244, opened 2023-10-26
Add dockerfile

dtrifiro:
- Add a `Dockerfile` definition: this copies the built wheel into the container and installs it
- Add a github actions workflow to build the dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,43 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest as builder
+
+RUN microdnf update -y && \
+    microdnf install -y \
+        git python39-pip && \
+    pip3 install --upgrade --no-cache-dir pip && \
+    microdnf clean all
+
+RUN python3 -m venv /opt/caikit/
+
+ENV VIRTUAL_ENV=/opt/caikit
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+COPY dist/caikit_nlp*.whl /tmp/
+RUN pip install --no-cache /tmp/caikit_nlp*.whl && rm /tmp/caikit_nlp*.whl
+
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest as deploy
+
+RUN microdnf update -y && \
+    microdnf install -y \
+        shadow-utils python39 && \
+    microdnf clean all
+
+COPY --from=builder /opt/caikit /opt/caikit
+
+RUN groupadd --system caikit --gid 1001 && \
+    adduser --system --uid 1001 --gid 0 --groups caikit \
+    --home-dir /caikit --shell /sbin/nologin \
+    --comment "Caikit User" caikit
+
+ENV VIRTUAL_ENV=/opt/caikit
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+USER caikit
+
+ENV RUNTIME_LIBRARY=caikit_nlp
+# Optional: use `CONFIG_FILES` and the /caikit/ volume to explicitly provide a configuration file and models
+# ENV CONFIG_FILES=/caikit/caikit.yml
+VOLUME ["/caikit/"]
+WORKDIR /caikit
+
+CMD ["python"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,8 @@ RUN microdnf update -y && \
     microdnf clean all
 
 COPY --from=builder /opt/caikit /opt/caikit
+COPY LICENSE /opt/caikit/
+COPY README.md /opt/caikit/
 
 RUN groupadd --system caikit --gid 1001 && \
     adduser --system --uid 1001 --gid 0 --groups caikit \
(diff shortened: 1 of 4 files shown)


markstur (commented):
This is wonderful.  Thank you @dtrifiro !!

gkumbhat (commented):
Hi @dtrifiro , Thanks for adding the dockerfile and updates to the readme. The "official" version of caikit image is getting built using [this](https://github.com/opendatahub-io/caikit-tgis-serving/blob/main/Dockerfile) Dockerfile, which is maintained by RH teams. From the perspective of reducing duplicity and maintenance overhead, what is the plan from OpenData hub perspective for the Dockerfile over there vs building the image in this repo ?

Xaenalt:
So, the overall thinking here is that it's better to have stuff in the upstream, this would let any users find it easier and work with it easier, and so the ODH ones would migrate upstream and we'd just be pulling it down for the RHODS builds. This would also let the upstream build have rolling dependencies, whereas they could be fixed downstream on a release cadence

I think an ideal path is to do more of the development here in caikit's upstream, so that we're not carrying large downstream deltas, what do you think @gkumbhat ?

dtrifiro:
Hey @gkumbhat sorry to always bug you :sweat_smile:, but is there anything I can do to help get this merged?

Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+USER caikit
+
+ENV RUNTIME_LIBRARY=caikit_nlp
+# Optional: use `CONFIG_FILES` and the /caikit/ volume to explicitly provide a configuration file and models
+# ENV CONFIG_FILES=/caikit/caikit.yml
+VOLUME ["/caikit/"]
+WORKDIR /caikit
+
+CMD ["python",  "-m", "caikit.runtime"]
```
gkumbhat, line 43: Since executing through runtime isn't main purpose of this repo, lets remove `caikit.runtime` as the entry point here.

dtrifiro: Anything else you propose to use as `CMD` or should we just leave this empty, making `bash` the default command?

README.md
```diff
@@ -18,8 +18,122 @@ Capabilities provided by `caikit-nlp`:
...
+  "input_token_count": 10,
+  "seed": null
+}
+```
+
+All the available API endpoints and protos can be dumped using [`scripts/dump_apis.sh`](/scripts/dump_apis.sh).
+
+#### Docker
+
+`caikit-nlp` can also be deployed as a Docker container. To build the docker image:
```
gkumbhat, line 108: as per my suggestion above regarding entrypoint, we would also need to modify this section a bit.

Dockerfile
gkumbhat, line 1: Can we add license, description etc to this dockerfile that way a user have those information in the container automatically

gkumbhat (changes requested):

README.md
dtrifiro: Done

Dockerfile
dtrifiro: Added `LICENSE` and `README.md`.

We also have those for free by installing the wheel package:

```console
$ ls -/opt/caikit/lib/python3.9/site-packages/caikit_tgis_backend-0.1.21.dist-info:
INSTALLER  LICENSE  METADATA  RECORD  WHEEL
```
`METADATA` contains  package metadata and the embedded readme

README.md
```diff
@@ -18,8 +18,122 @@ Capabilities provided by `caikit-nlp`:
...
+```
+
+Then convert it:
+
+```python
+import caikit_nlp
+
+model_path="flan-t5-small"
+model = caikit_nlp.text_generation.TextGeneration.bootstrap(model_path)
+model.save(f"{model_path}-caikit")
```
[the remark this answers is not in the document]
dtrifiro: Done

README.md
```diff
@@ -16,20 +16,142 @@ Capabilities provided by `caikit-nlp`:
...
+### Bootstrapping models
+
+`caikit_nlp` can use  Hugging Face models, allowing for direct download and bootstrapping.
+
+For example, to use [google/flan-t5-small](https://huggingface.co/google/flan-t5-small):
+
+```python
+import os
+# The env var ALLOW_DOWNLOADS has to be set to allow model downloads before importing caikit_nlp
+os.environ['ALLOW_DOWNLOADS'] = "1"
```
gkumbhat, line 44: there are some other ways to configure this as well, like via config file, but I think keeping this referring to environment variable is more helpful that way a user instantly knows that they can just configure this via environment variable. No change needed, but just mentioning it here.

README.md
```diff
@@ -16,20 +16,142 @@ Capabilities provided by `caikit-nlp`:
...
+  "generated_tokens": 5,
+  "finish_reason": "MAX_TOKENS",
+  "producer_id": {
+    "name": "Text Generation",
+    "version": "0.1.0"
+  },
+  "input_token_count": 10,
+  "seed": null
+}
+```
```
gkumbhat, line 98: Later on we can also move these into the `examples` folder.. We can have 2 parts of the "examples", 1 related to notebooks etc and other related to servers including examples from here 🤔

gkumbhat (approved):
LGTM. Thank you

PR #258, opened 2023-11-01
:bug: Fix loss tensor to float conversion

gkumbhat:
### Changes
- Change how we fetch the float value from loss tensor to avoid potentially weird memory issues

Ref blog talking about this: https://medium.com/@raghadalghonaim/memory-leakage-with-pytorch-23f15203faa4


alex-jw-brooks (approved):
This is pretty sinister, can you link that blog we were looking at in case we run into it in the future on accident?

LGTM!

PR #259, opened 2023-11-02
:goal_net: Disallow empty train streams

evaline-ju:
Re-creation of #223 - this had to be reverted because of https://github.com/caikit/caikit/issues/524, which has since been addressed via https://github.com/caikit/caikit/pull/553 and released in caikit 0.23.1

Addresses https://github.com/caikit/caikit-nlp/issues/214


gkumbhat (approved):
LGTM.

Issue #257, opened 2023-11-01
inference after fine tuning llama-7b fails


Ssukriti:
## Describe the bug

after fine tuning the llama-7b-model , on running inference I see 
`"/dccstor/ssharma/caikit-nlp/examples/run_fine_tuning.py", line 347, in <module>
  print("Generated text: ", prediction_results)
 File "/dccstor/ssharma/caikit_nlp_env_new/lib/python3.9/site-packages/caikit/core/data_model/base.py", line 1004, in __repr__
  return [self.to](http://self.to/)_json(indent=2, ensure_ascii=False)
 File "/dccstor/ssharma/caikit_nlp_env_new/lib/python3.9/site-packages/caikit/core/data_model/base.py", line 987, in to_json
  return json.dumps([self.to](http://self.to/)_dict(), **kwargs)
 File "/dccstor/ssharma/caikit_nlp_env_new/lib/python3.9/site-packages/caikit/core/data_model/base.py", line 946, in to_dict
  dict_value = self._field_to_dict_element(field)
 File "/dccstor/ssharma/caikit_nlp_env_new/lib/python3.9/site-packages/caikit/core/data_model/base.py", line 1032, in _field_to_dict_element
  enum_rev[attr.value] if isinstance(attr, Enum) else enum_rev[attr]
KeyError: 'MAX_TOKENS'``

## Platform

Please provide details about the environment you are using, including the following:
- Library version: latest

## Sample Code

running examples/fine_tuning.py on llama-7b-model results in thsi failure on `print(model.run())` after fine tuning

## Expected behavior

Infrence should run and output should be printed

## Observed behavior
`"/dccstor/ssharma/caikit-nlp/examples/run_fine_tuning.py", line 347, in <module>
  print("Generated text: ", prediction_results)
 File "/dccstor/ssharma/caikit_nlp_env_new/lib/python3.9/site-packages/caikit/core/data_model/base.py", line 1004, in __repr__
  return [self.to](http://self.to/)_json(indent=2, ensure_ascii=False)
 File "/dccstor/ssharma/caikit_nlp_env_new/lib/python3.9/site-packages/caikit/core/data_model/base.py", line 987, in to_json
  return json.dumps([self.to](http://self.to/)_dict(), **kwargs)
 File "/dccstor/ssharma/caikit_nlp_env_new/lib/python3.9/site-packages/caikit/core/data_model/base.py", line 946, in to_dict
  dict_value = self._field_to_dict_element(field)
[...]

PR #247, opened 2023-10-27
fix uncaught exception in http client

dtrifiro:
fixes #245

PR #261, opened 2023-11-02
fix:finish code to TextGeneration object

Ssukriti:
Closes: https://github.com/caikit/caikit-nlp/issues/257


evaline-ju (commented):
I did notice https://github.com/caikit/caikit-nlp/pull/247 addressed the same issue (though it does not add a test)

PR #262, opened 2023-11-03
tests: make models fixtures session-scoped

dtrifiro:
Slightly improve tests times by marking some expensive fixtures as session scoped


gkumbhat (approved):
LGTM

Issue #263, opened 2023-11-06
Update peft dependency to 0.6.0 release


evaline-ju:
## Description

The [latest release of peft](https://github.com/huggingface/peft/releases/tag/v0.6.0) includes MPT. Using this instead of the git hash dependency will allow us to make pypi releases of `caikit-nlp`. This will also allow anything that depends on `caikit-nlp` to be pulled in via version rather than git hash.

## Acceptance Criteria

- [ ] Update git hash dependency to version dependency
- [ ] Make pypi release of `caikit-nlp`
- [ ] No functionalities are broken with the upgrade

PR #265, opened 2023-11-07
deps: update peft to install from pypi

anhuong:
relates to: #263

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,11 +28,7 @@ dependencies = [
     "torch>=2.0.1",
     "tqdm>=4.65.0",
     "transformers>=4.32.0",
-    # GK-AUG-25-2023 NOTE: mpt branch on Mayank's fork was merged to peft main on Aug 24 and it got deleted
-    # which broke caikit-nlp build. peft hasn't released newer version yet, so to get
-    # the build fix, we pulling peft from main branch commit. In future, we will pull PEFT from
-    # pypi
-    "peft@git+https://github.com/huggingface/peft.git@8c17d556a8fe9522e10d73d7bd3fad46a6ecae14"
+    "peft>=0.6.0",
 ]
 
 [tool.setuptools.packages.find]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     "torch>=2.0.1",
     "tqdm>=4.65.0",
     "transformers>=4.32.0",
-    "peft>=0.6.0",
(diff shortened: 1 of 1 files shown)


anhuong:
I tested this by running the tox unit tests and by running the `Getting_Started` jupyter notebook. I haven't worked with NLP too much, is there another way that this should be tested? Also, it looks like that on the next release tag it will trigger the pypi push - does this need to be tested?

gkumbhat:
@anhuong thanks for fixing the package here @anhuong . We would need to do an end to end training test with this, to make sure the training goes through smoothly for seq2seq and causal-lm models and also we are able to do inferencing on the tuned prompts. 🤔

pyproject.toml
```diff
@@ -28,11 +28,7 @@ dependencies = [
     "torch>=2.0.1",
     "tqdm>=4.65.0",
     "transformers>=4.32.0",
-    # GK-AUG-25-2023 NOTE: mpt branch on Mayank's fork was merged to peft main on Aug 24 and it got deleted
-    # which broke caikit-nlp build. peft hasn't released newer version yet, so to get
-    # the build fix, we pulling peft from main branch commit. In future, we will pull PEFT from
-    # pypi
-    "peft@git+https://github.com/huggingface/peft.git@8c17d556a8fe9522e10d73d7bd3fad46a6ecae14"
+    "peft>=0.6.0",
```
gabe-l-hart, line 31: I think it's safer to guard against breaking changes. More annoying to maintain, but safer

```suggestion
    "peft>=0.6.0,<0.7.0",
```

gkumbhat (approved):
LGTM

gkumbhat:
@anhuong actually lets go with @dtrifiro 's idea of pinning PEFT. That would be a more resilient way of handling for this case.

gkumbhat (changes requested):
Lets go with @dtrifiro's suggestion of pinning peft, that would be a bit more resilient to sudden or surprise breakage

evaline-ju:
Verified training ability and inference after deployment  (PT, MPT on causal lm [bloom] and seq2seq [flan-t5-xl])

Due to small training differences with the HF trainer (on the latest changes `0.4.0`/current main) I verified that results were consistent on inference for the same models

With only a peft change (I.e. no rebase/latest caikit-nlp updates, off `0.3.3` changes) I was able to confirm same results from two prompt tuned causal-lm (bloom) models

evaline-ju (approved):
LGTM

Issue #243, opened 2023-10-25
.run() in local text generation block needs flag `preserve_input_text` 


Ssukriti:
## Is your feature request related to a problem? Please describe.

Since prompt tuning of causal LMs cause input text to be repeated with prediction generated, .run() for local text generation needs a flag `preserve_input_text`  to optionally get rid of input text when set to False. This will improve the quality of running local inferencing using caikit-nlp

## Describe the solution you'd like
`preserve_input_text` added to .run() signature https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/modules/text_generation/text_generation_local.py#L521 so that input text can be removed from output

Ssukriti:
PR: https://github.com/caikit/caikit-nlp/pull/267 waiting for review

Ssukriti:
merged and closing

PR #267, opened 2023-11-09
Add preserve input text inference

Ssukriti:
for: https://github.com/caikit/caikit-nlp/issues/243

adds a flag to preserve or remove input text from output of CAUSAL models (**local inferencing** of prompt tuning and generation)

this will achieve functional parity with TGIS

The change will enable local evaluation of causal models without TGIS

### DIFF
--- a/caikit_nlp/toolkit/text_generation/model_run_utils.py
+++ b/caikit_nlp/toolkit/text_generation/model_run_utils.py
@@ -152,6 +152,7 @@ def generate_text_func(
         Union[Tuple[int, float], ExponentialDecayLengthPenalty]
     ] = None,
     stop_sequences: Optional[List[str]] = None,
+    preserve_input_text: bool = True,
     **kwargs,
 ):
     """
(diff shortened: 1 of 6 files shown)


caikit_nlp/toolkit/text_generation/model_run_utils.py
```diff
@@ -251,14 +266,37 @@ def generate_text_func(
...
         input_token_count=input_token_count,
         seed=seed,
     )
 
 
+def __postprocess_remove_input_text(tokenizer, preds, inputs, task_type):
+    """For Causal LM task types, preserve_input_text set to False will
+    remove the input text from generated output.
+    """
+    if task_type == "CAUSAL_LM":
```
Ssukriti, line 281: HF pipeline has implemented this for Causal LMs generation pipelines https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/text_generation.py#L287 

I took the same logic as it was suggested by other stack overflow posts too. I set the flag to True by default as generate() returns full text by default. 

I have tested this with fine tuning on llama model with different texts and found it to work as expected. Unit tests show the same

Ssukriti:
actually hold on, making some changes

Ssukriti:
@gkumbhat its ready now when you have a chance, thanks

gkumbhat (approved):
LGTM. Thanks Sukriti

PR #268, opened 2023-11-17
:bug: Fix the ordering of arguments in all the stream text gen functions

gkumbhat:
### Changes
- Fix run function ordering issue


gabe-l-hart (approved):
Looks good, thanks for catching this!

Issue #136, opened 2023-08-21
Add support for LoRA in prompt tuning module (Investigate)


gkumbhat:
## Is your feature request related to a problem? Please describe.

Currently we only support vanilla prompt tuning and multi-task prompt tuning from PEFT library. LoRA has shown to give good performance on variety of datasets. This story is to investigate support for LoRA into the module.

chakrn:
Should this be added to the 'Add Support for LoRA' milestone? cc: @rawkintrevo

ibm-peach-fish:
@chakrn / @gkumbhat I think we can close this as "no" i did the second iteration of adding LoRA in Prompt Tuning and it was pushed back.

chakrn:
cc: @Ssukriti

Ssukriti:
doing this in steps, but first PR ready for review https://github.com/caikit/caikit-nlp/pull/270 @gkumbhat

PR #270, opened 2023-11-17
Lora config data model and utilities

Ssukriti:
For #https://github.com/caikit/caikit-nlp/issues/136

This PR is the first step to adding a data model and utility to initialize LoraConfig . It also includes a minor refactor to move common config initialization functions between PEFT and LORA


caikit_nlp/modules/text_generation/peft_config.py
```diff
@@ -218,3 +196,78 @@ def get_peft_config(
...
+    log.info(
+        "<NLP18184771I>",
+        "[{}] config params not supported by provided tuning type!".format(
+            params.keys() - allowed_params.keys()
+        ),
+    )
+    return allowed_params
+
+
+def get_lora_config(tuning_type, tuning_config, base_model) -> LoraConfig:
```
Ssukriti, line 256: I created a new function that can be called by .train() of any module. most of get_peft_config was specific to PEFTConfig . I moved the common functions to  independent functions

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -1150,34 +1155,6 @@ def _execute_train_loop(
                     )
         return {"loss": training_loss_tracker}
 
-    @classmethod
-    def _filter_params_for_prompt_config(cls, prompt_config, params):
-        """Utility function to filter out required parameters for prompt_config
```
Ssukriti, line 1155: I think this belongs in peft_config as it a utility to create any Config. I also think `create_hf_tuning_config(` in thsi file belongs there, but I did not move it because I dont know the rational behind heeping `create_hf_tuning_config` here in prompt_tuning.py

caikit_nlp/data_model/generation.py
```diff
@@ -73,6 +73,41 @@ class TuningConfig(DataObjectBase):
...
+    # it will apply the LoRA transformations
+    # on the layer indexes that are specified in this list.
+    # If a single integer is passed,
+    # it will apply the LoRA transformations on the layer at this index.
+    layers_to_transform: Union[List[int], int]
+    # The layer pattern name, used only if layers_to_transform is different from None and
+    # if the layer pattern is not in the common layers pattern.
+    layers_pattern: str
+    output_model_types: List[str]
+
```
Ssukriti, line 109: I didnt add few  parameters from LoraConfig that are less frequently used. Everything that we will want to tweak is added

caikit_nlp/modules/text_generation/peft_config.py
```diff
@@ -96,12 +96,128 @@ def resolve_base_model(base_model, cls, torch_dtype):
...
+    error.value_check(
+        "<NLP83837412E>",
+        is_valid_verbalizer(verbalizer),
+        "Provided verbalizer is an invalid type or has no renderable placeholders",
+    )
+
+    # Coerce the passed model into a resource; if we have one, this is a noop
+    # TODO: When splitting up this mono-module, use the configured resource
+    #   type of the concrete class to bootstrap
+    torch_dtype = get_torch_dtype(torch_dtype)
```
Ssukriti, line 145: torch_dtype is not even used, didnt remove it as it would be an API breaking change, but let me know if its ok to remove here

caikit_nlp/modules/text_generation/peft_config.py
```diff
@@ -96,12 +96,128 @@ def resolve_base_model(base_model, cls, torch_dtype):
...
+
+    error.value_check(
+        "<NLP30542004E>",
+        len(output_model_types) <= base_model.MAX_NUM_TRANSFORMERS,
+        f"Too many output model types. Got {len(output_model_types)}, "
+        f"maximum {base_model.MAX_NUM_TRANSFORMERS}",
+    )
+    # Ensure that our verbalizer is a string and will not render to a hardcoded string
+    error.value_check(
+        "<NLP83837412E>",
```
Ssukriti, line 137: verbalizer is also not used, we are checking if its a valid verbalizer , but that should be done in train of peft_prompt_tuning. it has nothing to do with the Config, so let me know if ok to remove (API change again)

caikit_nlp/modules/text_generation/peft_config.py
```diff
@@ -216,5 +276,15 @@ def get_peft_config(
         tuning_config=tuning_config,
         output_model_types=output_model_types,
     )
+    return peft_config
```
Ssukriti, line 279: this `create_hf_tuning_config` function should ideally be moved here. I dont know rationale for why its in peft_prompt_tuning.py . Let me know if we can move it here and then get rid of `cls` argument

caikit_nlp/data_model/generation.py
```diff
@@ -73,6 +73,41 @@ class TuningConfig(DataObjectBase):
     # encoder_hidden_size: int # Optional -  The hidden size of the prompt encoder.
 
 
+@caikit.core.dataobject(package="caikit_data_model.caikit_nlp")
+class LoraTuningConfig(DataObjectBase):
+    # Lora attention dimension.
+    r: int
+    # The names of the modules to apply Lora to.
+    target_modules: Union[List[str], str]
```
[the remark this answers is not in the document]
Ssukriti: I added the same from HF. There are ways to know modules of a model by loading model with transformers and printing it. I will have to include that in actual examples that we commit to caikit NLP when feature is complete. we might have to expose a helper function in caikit NLP to get_module_names(model) or something . but I think I can implement that in follow up PRs when we get to adding examples.

caikit_nlp/data_model/generation.py
```diff
@@ -73,6 +73,41 @@ class TuningConfig(DataObjectBase):
...
+    # The layer indexes to transform, if this argument is specified,
+    # it will apply the LoRA transformations
+    # on the layer indexes that are specified in this list.
+    # If a single integer is passed,
+    # it will apply the LoRA transformations on the layer at this index.
+    layers_to_transform: Union[List[int], int]
+    # The layer pattern name, used only if layers_to_transform is different from None and
+    # if the layer pattern is not in the common layers pattern.
+    layers_pattern: str
+    output_model_types: List[str]
```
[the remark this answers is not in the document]
Ssukriti: removed the above mentioned ones

caikit_nlp/modules/text_generation/peft_config.py
Ssukriti: I added deprecation warning

caikit_nlp/modules/text_generation/peft_prompt_tuning.py
alex-jw-brooks: It's more of a legacy reason than anything, all of the code for tuning config stuff was originally written in this file as part of this module, and the refactoring to pull some of it out happened later. I don't think there was a deep reason for leaving `create_hf_tuning_config` here

caikit_nlp/modules/text_generation/peft_config.py
```diff
@@ -96,12 +96,142 @@ def resolve_base_model(base_model, cls, torch_dtype):
 
 
 def get_peft_config(
-    tuning_type, tuning_config, base_model, cls, torch_dtype, verbalizer
+    tuning_type,
+    tuning_config,
+    base_model,
+    cls=None,
+    torch_dtype=None,
+    verbalizer="{{input}}",
```
alex-jw-brooks, line 104: is there a reason these defaults are now being set here? I.e., the calling module is always going to pass them in positionally and override these defaults, right?

caikit_nlp/modules/text_generation/peft_config.py
```diff
@@ -208,13 +282,23 @@ def get_peft_config(
...
 
-    return task_type, output_model_types, peft_config, tuning_type
+
+def _create_lora_config(tuning_config, task_type) -> LoraConfig:
+    """Creates Huggingface LoraConfig from Caikit tuning configuration."""
+
+    config_kwargs = tuning_config.to_dict()
+    log.info("<NLP61012781I>", f"Parameters used: {config_kwargs}")
+    config_params = _filter_params_for_prompt_config(tuning_config, config_kwargs)
+    del config_params["output_model_types"]
```
alex-jw-brooks, line 302: I am probably missing something - what is the reason for deleting this?

alex-jw-brooks (commented):
In general I think this looks good! Some small questions, but once those are cleared up I think we can get it merged, thanks Sukriti!

caikit_nlp/data_model/generation.py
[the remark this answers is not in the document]
gkumbhat: We would also need to figure out how do we expose these functions to cloud users then, also we would need to explain them what these modules are and what are its implication. Any suggestions on how this would be exposed to cloud users?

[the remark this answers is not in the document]
Ssukriti: also, when users have their own model they will most likely know the architecture

PR #272, opened 2023-11-20
Bump caikit to 0.25

markstur:


markstur:
Now we cannot update until because embedding interfaces are in both.  Need to update with reranker PR.  #235

PR #274, opened 2023-11-20
Text Embedding: Add optimizations with PT2_COMPILE and/or IPEX

markstur:
Add flag to use pytorch compile and/or try IPEX.

To use IPEX:
1.  Install intel-extension-for-pytorch
2. Set env var USE_IPEX=true

Default is false.  If set to true, but the import fails, then a warning will be logged and USE_IPEX will be disabled.
Optionally, USE_XPU may be set to true to allow IPEX to use its xpu (only applies when IPEX is active).

USE_MPS has been added for GPU testing on Macbook (ignored when IPEX is used).

To invoke torch.compile() optimization, set the env var PT2_COMPILE=true.
torch.compile() is not available on all versions/platforms.  A warning message is logged if this is enabled and throws an exception.

### DIFF
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -17,6 +17,10 @@ from typing import List, Optional
 import importlib
 import os
 
+# Third Party
+from torch.backends import mps
+import torch
+
 # First Party
 from caikit.core import ModuleBase, ModuleConfig, ModuleSaver, module
 from caikit.core.data_model.json_dict import JsonDict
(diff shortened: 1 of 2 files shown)


evaline-ju (commented):
source code updates look good to me in theory but there are a decent number of code branches - would it be possible to add a few small unit test cases to check that the expected device is used/set with any configuration?

markstur:
> source code updates look good to me in theory but there are a decent number of code branches - would it be possible to add a few small unit test cases to check that the expected device is used/set with any configuration?

Yep. The way it was written, I just didn't think unit tests made sense (which I know obviously meant it was written wrong).  So I refactored and added tests to cover most of the code.  Not mocking the actual optimize or compile at this point, but enough tests to get through those decision branches with monkeypatch env var fixtures.

Also added one line to "warmup" the model.

evaline-ju (approved):
LGTM! Thanks for adding tests

PR #280, opened 2023-11-27
Add max size information to error message

mynhardtburger:
For user understandability include the actual and max training sizes in the error message.


gkumbhat (approved):
LGTM

gkumbhat (changes requested):
@mynhardtburger we actually need 2 PRs here:
1. based on `main` and towards `main`
2. based on `release-0.3` and towards `release-0.3`

The later one (2) is what we would use to create a newer release `0.3.2`

mynhardtburger:
@gkumbhat. PR has been rebased against main.

evaline-ju (approved):
LGTM

PR #289, opened 2023-12-01
dockerfile: build wheel in container

dtrifiro:
Dockerfile was consuming a previously built wheel because setuptools_scm requires access to the git repository. By mounting the `.git` directory at the build step, we can build in the container.


Xaenalt (approved):

gkumbhat (approved):
LGTM

PR #292, opened 2023-12-07
Add data collator for causal completion training

Ssukriti:
This efforts has ability for `completion only LM` Training as an optional flag. How it should be exposed to users on product is being discussed, but we want to add the ability to support it in library for **Causal LMs**

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
+++ b/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
@@ -350,7 +350,9 @@ class PeftPromptTuning(ModuleBase):
             seed: int
                 Integer to be used as random seed for training.
             train_on_completion: bool
-                True will train the model on the generated prompts only. Default: False.
+                True will train the model on the generated prompts only.
+                Only applicable to Causal LMs.
+                Default: False.
             response_template: Optional[str] = None
                 Only if train_on_completion is set to True, pass a response template that
                 will be used to parse out the response.
--- a/caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
+++ b/caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
@@ -25,6 +25,7 @@ from transformers import (
     DataCollatorForLanguageModeling,
 )
 from transformers.models.auto import modeling_auto
+from trl import DataCollatorForCompletionOnlyLM
 import torch
 
 # First Party
@@ -168,6 +169,18 @@ class HFAutoCausalLM(PretrainedModelBase):
                 Collator to be used for causal language modeling.
         """
 
+        if "train_on_completion" in kwargs and kwargs["train_on_completion"]:
+            applicable_args = ["mlm", "response_template", "instruction_template"]
+            collator_kwargs = {
+                key: kwargs[key] for key in applicable_args if key in kwargs
+            }
+
+            if "mlm" not in collator_kwargs:
+                collator_kwargs["mlm"] = False
+
+            return DataCollatorForCompletionOnlyLM(
+                tokenizer=self._tokenizer, return_tensors="pt", **collator_kwargs
+            )
         applicable_args = ["mlm", "pad_to_multiple_of"]
         collator_kwargs = {key: kwargs[key] for key in applicable_args if key in kwargs}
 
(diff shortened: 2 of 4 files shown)


caikit_nlp/modules/text_generation/peft_prompt_tuning.py
```diff
@@ -300,6 +300,8 @@ def train(
         torch_dtype: Optional[str] = None,  # TODO: Optional[Union[torch.dtype, str]]
         silence_progress_bars: Optional[bool] = True,
         seed: int = RANDOM_SEED,
+        train_on_completion: bool = False,
+        response_template: str = None,
```
Ssukriti, line 304: decision to expose this flag may change, just fYI. I wont be merging this code right away, but I want to ensure rest of it looks ok, assuming it is a user provided argument

caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
```diff
@@ -168,6 +169,18 @@ def _get_data_collator(self, **kwargs) -> "transformers.DataCollator":
...
+            applicable_args = ["mlm", "response_template", "instruction_template"]
+            collator_kwargs = {
+                key: kwargs[key] for key in applicable_args if key in kwargs
+            }
+
+            if "mlm" not in collator_kwargs:
+                collator_kwargs["mlm"] = False
+
+            return DataCollatorForCompletionOnlyLM(
+                tokenizer=self._tokenizer, return_tensors="pt", **collator_kwargs
```
alex-jw-brooks, line 182: Have you tested this with multiGPU by any chance? In the past, we've seen some kind of strange behavior for some of the collators around dynamic padding, curious if that's been observed here

caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
```diff
@@ -168,6 +169,18 @@ def _get_data_collator(self, **kwargs) -> "transformers.DataCollator":
...
+        if "train_on_completion" in kwargs and kwargs["train_on_completion"]:
+            applicable_args = ["mlm", "response_template", "instruction_template"]
+            collator_kwargs = {
+                key: kwargs[key] for key in applicable_args if key in kwargs
+            }
+
+            if "mlm" not in collator_kwargs:
+                collator_kwargs["mlm"] = False
+
+            return DataCollatorForCompletionOnlyLM(
```
alex-jw-brooks, line 181: The SFT Trainer hasn't been integrated quite yet, correct? Could you add a comment here for a TODO to validate that this can't be used if the trainer is initialized with `packing=True` so that we don't miss that edge case in the future? https://huggingface.co/docs/trl/v0.7.4/en/sft_trainer#train-on-completions-only

alex-jw-brooks (changes requested):
I think this looks good sukriti! Some thoughts, thanks!

caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
Ssukriti: Raghu's team has tested it with MultiGPU, but the plan is to only use this codebase for single GPU going forward. MultiGPu will go through non caikit path

pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
     "torch>=2.0.1",
     "tqdm>=4.65.0",
     "transformers>=4.32.0",
+    "trl>=0.7.2",
```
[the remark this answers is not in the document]
Ssukriti: oh yes, thanks, will do

examples/run_peft_tuning.py
```diff
@@ -241,6 +241,18 @@ def register_common_arguments(subparsers: Tuple[argparse.ArgumentParser]) -> Non
             default="float32",
             choices=["float16", "bfloat16", "float32"],
         )
+        subparser.add_argument(
+            "--train_on_completion",
+            help="Train on completion True or False",
+            default=False,
+            type=bool,
+            choices=[True, False],
```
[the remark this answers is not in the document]
Ssukriti: thanks, I didnt even mean to commit this example actually, I will make this change and test that example actually works.

caikit_nlp/resources/pretrained_model/hf_auto_causal_lm.py
Ssukriti: background being we wnat to enable completion training for current PT and LOra which will be through thsi codebase

PR #304, opened 2024-01-17
:goal_net: Handle empty text for filtered span classification

evaline-ju:
Return empty classification results when classification is invoked with empty text or empty text stream. This would prevent potential further errors that _may_ incur if any models used (tokenizers or classifiers cannot handle empty strings properly). Empty text _could_ have been considered an error but based on current usage this would not be a desired user experience

### DIFF
--- a/caikit_nlp/modules/token_classification/filtered_span_classification.py
+++ b/caikit_nlp/modules/token_classification/filtered_span_classification.py
@@ -141,7 +141,7 @@ class FilteredSpanClassification(ModuleBase):
 
         if threshold is None:
             threshold = self.default_threshold
-        if text == "":
+        if not text:
             # Allow empty text case to fall through - some tokenizers or
             # classifiers may error on this
             return TokenClassificationResults(results=[])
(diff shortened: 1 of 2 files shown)


caikit_nlp/modules/token_classification/filtered_span_classification.py
```diff
@@ -136,8 +136,16 @@ def run(
         Returns:
             TokenClassificationResults
         """
+        error.type_check("<NLP82129006E>", str, text=text)
+        error.type_check("<NLP01414077E>", float, allow_none=True, threshold=threshold)
+
         if threshold is None:
             threshold = self.default_threshold
+        if text == "":
```
gkumbhat, line 144: is there a reason to not throw error on this? is that for compatibility with TGIS?

gabe-l-hart: NIT: Is it worth just using the native "truthiness" of `text` here (`if not text`)?

gkumbhat (approved):
LGTM

PR #308, opened 2024-01-25
Add model load logging

gkumbhat:
### Changes
- Add `platformdirs` for formatter to avoid errors on mac for `black`
- Add time based context logging for tokenizer and model loading for base models. This will allow us to get better picture of training time.


evaline-ju (approved):
LGTM!

PR #316, opened 2024-02-03
Embeddings: Fix retry error handling to return first exception. Default to zero retries.

### LINKED ISSUE #315: Embedding retry returns the wrong exception
## Describe the bug

The retry (which was added to mitigate thread contention "already borrowed" errors) had some undesirable behavior. It was hiding actual errors. Also, now that we got that original problem fixed, this generally should be turned off so we won't eat-and-repeat on real errors.

## Expected behavior

1.  If enabled and no success after the configured number of retries, then return the original exception instead of the one that says "too many retries".  Use logging for the retry attempts.

2.  Default to zero retries because this is ideally no longer used (but can be enabled by env var if something comes up again).

## Observed behavior

When actual errors that repeatedly failed happened, the exception was RuntimeError("Too many retries...") instead of a more usable error such as an ValueError.

markstur:
The retry (which was added to mitigate thread contention "already borrowed" errors) had some undesirable behavior. It was hiding the actual error. Also, now that we got that original problem fixed, this generally should be turned off so we won't eat-and-repeat on real errors.

* Default to no retries
* Capture the original exception and raise that if retries don't succeed (log the retries)
* Make it more testable and test the exception handling better.

Fixes #315

### DIFF
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -248,7 +248,7 @@ class EmbeddingModule(ModuleBase):
                     first_exception = e
                 if self.RETRY_COUNT > 0:
                     warn_msg = f"Try {count + 1}: {fn} failed due to: {e}"
-                    logger.warning("<NLP54902271E>", warn_msg, exc_info=True)
+                    logger.warning("<NLP54902271W>", warn_msg, exc_info=True)
                     if count + 1 < self.RETRY_COUNT:
                         time.sleep(0.1 * (count * 2))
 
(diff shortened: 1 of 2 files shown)


caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -238,17 +238,25 @@ def _optimize(model, ipex, device):
...
             except Exception as e:  # pylint: disable=broad-exception-caught
-                warn_msg = f"Retry {fn} due to: {e}"
-                logger.warning(warn_msg, exc_info=True)
-                time.sleep(0.1 * (count * 2))
-        error.log_raise("<NLP31069292E>", RuntimeError(f"Too many retries of fn={fn}"))
+                if first_exception is None:
+                    first_exception = e
+                if self.RETRY_COUNT > 0:
+                    warn_msg = f"Try {count + 1}: {fn} failed due to: {e}"
+                    logger.warning("<NLP54902271E>", warn_msg, exc_info=True)
```
evaline-ju, line 251: nit: code matching log level
```suggestion
                    logger.warning("<NLP54902271W>", warn_msg, exc_info=True)
```

evaline-ju (approved):
LGTM with nit

PR #319, opened 2024-02-07
Add sentence-transformers to dependencies for real

markstur:
Need to make sentence-transformers part of the default caikit-nlp runtime image. Previously this was an optional import requiring a special runtime image.

* Make sentence-transformers one of the dependencies
* Remove the just-for-testing no-deps support that was a work-around


evaline-ju (approved):
LGTM

PR #324, opened 2024-02-15
Adding tgis params to caikit api

waleedqk:
Add support for ReturnOptions, GeneratedTokens and InputTokens

As part of caikit there are some parameters that are hard-coded to default values in code.
In the TGIS implementation, users can set these options and get the appropriate response.
The parameters in questions are part of the ResponseOptions proto found [here](https://github.com/IBM/text-generation-inference/blob/main/proto/generation.proto).

This PR updates the data-model so that the return from inference server can be part of the result if the appropriate parameter is set in caikit FastAPI endpoint.


caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -202,6 +202,10 @@ def run(
         stop_sequences: Optional[List[str]] = None,
         seed: Optional[np.uint64] = None,
         preserve_input_text: bool = False,
+        input_tokens: bool = False,
+        generated_tokens: bool = True,
+        token_logprobs: bool = True,
+        token_ranks: bool = True,
```
evaline-ju, line 208: Was there a particular reason to set all of these to `True`? These all seem like extra response options that could be helpful for a user but might be extra overhead on computation

caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -40,6 +40,16 @@
...
         Whether or not the source string should be contained in the generated output,
         e.g., as a prefix.
+    input_tokens: str
+        Whether or not to include list of input tokens.  
+    generated_tokens: str
+        Whether or not to include list of individual generated tokens.  
+    token_logprobs: str
+        Whether or not to include logprob for each returned token.  
+        Applicable only if generated_tokens == true and/or input_tokens == true
+    token_ranks: str
```
evaline-ju, line 50: Based on the module runtime parameters these all appear to be `bool` not `str`? (Seems `preserve_input_text` might not have been accurate either, if that could also be updated here)

tests/fixtures/__init__.py
```diff
@@ -229,6 +237,10 @@ def stream_generate(request):
         token.text = "moose"
         token.logprob = 0.2
         fake_stream.tokens = [token]
+        input_tokens = mock.Mock()
+        input_tokens.text = "moose"
+        input_tokens.logprob = 0.2
+        fake_stream.input_tokens = [input_tokens]
```
evaline-ju, line 243: With the update of these text fixtures, could we add some assertions to the module tests that the new fields are passed along when expected (likely in the `validate_..` functions below in this file)?

evaline-ju (changes requested):
Thanks for the updates! A few questions/request for more test assertions

caikit_nlp/modules/text_generation/peft_tgis_remote.py
waleedqk: The main reason being the hard-coded values in `caikit_nlp/toolkit/text_generation/tgis_utils.py` were the very same. So I set the default here as was the current implementation.

evaline-ju: ah I see, thanks for the reminder. That's likely fine for parity then and user(s) can explicitly send `False` if needed

caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -37,9 +37,19 @@
...
         e.g., as a prefix.
+    input_tokens: bool
+        Whether or not to include list of input tokens.  
+    generated_tokens: bool
+        Whether or not to include list of individual generated tokens.  
+    token_logprobs: bool
+        Whether or not to include logprob for each returned token.  
+        Applicable only if generated_tokens == true and/or input_tokens == true
+    token_ranks: bool
+        Whether or not to include rank of each returned token.
```
evaline-ju, line 51: `token_ranks` is added but there is no output allowing for `rank` information to be exposed. Is this intentional?

waleedqk: yes, I had some trouble initially with figuring out the output for the `token_ranks` parameter.
Since it wasnt to core ask, I exposed the parameter but have not adjusted the output for it in the data-model or guardrails

evaline-ju (approved):
LGTM! Thanks for the update!

PR #328, opened 2024-02-27
Wrapped sentence transformer

markstur:
Wrapped sentence transformer to allow access to customized encode() and intercepting tokenizer() details.

* Needed for IPEX (Intel PyTorch extension) which needs bfloat16 and autocast settings to be effective.
* Allows us to get rid of the tokenizer copy that was needed for our desired truncation.  Now we leverage the same call SentenceTransformer was already doing anyway.

In order to keep the requested truncation error message, however, there is a re-tokenize just to get the correct error message.  This only happens on errors instead of impacting every request.

### DIFF
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -14,13 +14,16 @@
 
 # Standard
 from copy import deepcopy
-from typing import List, Optional
+from typing import List, Optional, Union
 import importlib
 import os
 import time
 
 # Third Party
 from torch.backends import mps
+from transformers.tokenization_utils_base import TruncationStrategy
+import grpc
+import numpy as np
 import torch
 
 # First Party
@@ -47,6 +50,7 @@ from caikit.interfaces.nlp.tasks import (
     SentenceSimilarityTask,
     SentenceSimilarityTasks,
 )
+from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
 import alog
 
 logger = alog.use_channel("TXT_EMB")
@@ -59,6 +63,7 @@ try:
     # Third Party
     from sentence_transformers import SentenceTransformer
     from sentence_transformers.util import (
+        batch_to_device,
         cos_sim,
         dot_score,
         normalize_embeddings,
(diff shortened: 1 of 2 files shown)


flaviabeo (approved):

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -90,6 +92,9 @@ def env_var_to_int(name, default):
 # Batch size for encode() if <= 0 or invalid, the sentence-transformers default is used
 BATCH_SIZE = env_var_to_int("BATCH_SIZE", default=0)
 
+# To use dtype=bfloat16 for IPEX with autocast in encode.
+BFLOAT16 = os.getenv("BFLOAT16", "false").lower() not in FALSY
```
evaline-ju, line 96: There is already a `torch_dtype` config - https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/config/config.yml#L19 which users can use as env var `TORCH_DTYPE` so I'm worried that continuing to add these variables as env vars could potentially be confusing, especially when multiple module use is considered.

Perhaps a config section related to embeddings and/or ipex could be considered if there's expected to be these sorts of variables. That way for use as env vars they would automatically be pre-pended with `EMBEDDINGS_` or `IPEX_` etc.

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -750,3 +654,172 @@ def save(self, model_path: str, *args, **kwargs):
...
+
+    def encode(
+        self,
+        sentences: Union[str, List[str]],
+        batch_size: int = 32,
+        device: str = None,
+        convert_to_numpy: bool = True,
+        convert_to_tensor: bool = False,
+        truncate_input_tokens: Optional[int] = 0,
+    ) -> np.ndarray:
```
evaline-ju, line 770: I think a docstring here could be helpful especially considering it's a non-private method

evaline-ju (commented):
A few comments for consideration

markstur:
There's a bug in the new config handling.  I'll use the DRAFT flag while I fix that.

tests/modules/text_embedding/test_embedding.py
```diff
@@ -699,7 +671,7 @@ def test_embeddings_with_truncation(truncate_input_tokens, loaded_model):
         repeat = truncate_input_tokens
 
     # Build a text like "x x x.. x " with room for one more token
-    repeat = repeat - 2  # space for start/end tokens
+    # repeat = repeat - 2  # space for start/end tokens
```
evaline-ju, line 674: nit: removable?

evaline-ju (commented):
I think the docstring request from last review is still remaining, otherwise LGTM

evaline-ju (approved):
LGTM!

PR #334, opened 2024-03-04
Include input token count in results

mynhardtburger:
Depends on caikit data model updates in this PR: https://github.com/caikit/caikit/pull/675

This extends the embedding module to include the `input_token_count` in the results of the `EmbeddingModule`'s `run_` methods.

The `sum_token_count(tokenized: BatchEncoding) -> int` function calculates the count of tokens requiring model attention, based on the `Encoding.attention_mask` property, as returned by `SentenceTransformerWithTruncate.tokenizer()`. [PAD] is irrelevant for truncation and max_token_count parameters, while [CLS] and [SEP] are counted by the model when it considers the max length and truncation.

Additionally tests to confirm sort order is maintained was added.

Various other quality of life type hints were added.

### DIFF
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -687,9 +687,10 @@ def get_sample_start_indexes(tokenized: BatchEncoding) -> List[int]:
     # note: tokenized["overflow_to_sample_mapping"] is a torch.Tensor
 
     samples_start_indexes: Dict[int, int] = {}
-    for i, sample in enumerate(tokenized["overflow_to_sample_mapping"]):
-        if int(sample) not in samples_start_indexes:
-            samples_start_indexes[int(sample)] = i
+    for i, tensor_sample in enumerate(tokenized["overflow_to_sample_mapping"]):
+        int_sample = int(tensor_sample)
+        if int_sample not in samples_start_indexes:
+            samples_start_indexes[int_sample] = i
 
     return list(samples_start_indexes.values())
 
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -842,7 +842,7 @@ class SentenceTransformerWithTruncate(SentenceTransformer):
         error.type_check_all("<NLP82314994E>", str, sentences=list_of_sentences)
 
         if device is None:
-            device = self._target_device
+            device = self.device
 
         self.to(device)
 
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -13,13 +13,15 @@
 # limitations under the License.
 
 # Standard
-from typing import List, Optional, Union
+from collections.abc import Iterable, Sized
+from typing import Callable, List, NamedTuple, Optional, Tuple, TypeVar, Union
 import importlib
 import os
 import time
 
 # Third Party
 from torch.backends import mps
+from transformers import BatchEncoding
 import numpy as np
 import torch
 
@@ -56,6 +58,7 @@ from caikit_nlp.modules.text_embedding.utils import env_val_to_bool, env_val_to_
 logger = alog.use_channel("TXT_EMB")
 error = error_handler.get(logger)
 
+
 # To avoid dependency problems, make sentence-transformers an optional import and
 # defer any ModuleNotFoundError until someone actually tries to init a model with this module.
 try:
@@ -87,6 +90,15 @@ RETRIES = env_val_to_int(val=embedding_cfg.get("retries"), default=0)
 BATCH_SIZE = env_val_to_int(val=embedding_cfg.get("batch_size"), default=0)
 DEVICE = embedding_cfg.get("device", "")
 
+RT = TypeVar("RT")  # return type
+
+
+class EmbeddingResultTuple(NamedTuple):
+    """Output of SentenceTransformerWithTruncate.encode()"""
+
+    embedding: np.ndarray
+    input_token_count: int
+
 
 @module(
     "eeb12558-b4fa-4f34-a9fd-3f5890e9cd3f",
@@ -102,7 +114,6 @@ DEVICE = embedding_cfg.get("device", "")
     ],
 )
 class EmbeddingModule(ModuleBase):
-
     # Retry count if enabled to try again (was for thread contention errors)
     RETRY_COUNT = max(RETRIES, 0)  # Ensure non-negative, before using in loop!
 
@@ -177,7 +188,7 @@ class EmbeddingModule(ModuleBase):
                     f"IPEX enabled in env, but skipping ipex.optimize() because "
                     f"import intel_extension_for_pytorch failed with exception: {ie}"
                 )
-                logger.warning(msg, exc_info=1)
+                logger.warning(msg, exc_info=True)
 
         return ret
 
@@ -210,7 +221,6 @@ class EmbeddingModule(ModuleBase):
 
     @staticmethod
     def _optimize(model, ipex, device, autocast, pt2_compile):
-
         if ipex:
             if autocast:  # IPEX performs best with autocast using bfloat16
                 model = ipex.optimize(
@@ -233,7 +243,7 @@ class EmbeddingModule(ModuleBase):
                 logger.warning(warn_msg, exc_info=True)
         return model
 
-    def _with_retry(self, fn, *args, **kwargs):
+    def _with_retry(self, fn: Callable[..., RT], *args, **kwargs) -> RT:
         first_exception = None
         for count in range(1 + self.RETRY_COUNT):  # try once plus retries (if needed)
             try:
@@ -253,7 +263,7 @@ class EmbeddingModule(ModuleBase):
             exception=first_exception,
         )
 
-    def _encode_with_retry(self, *args, **kwargs):
+    def _encode_with_retry(self, *args, **kwargs) -> EmbeddingResultTuple:
         """All encode calls should use this for consistent param adding and retry loop"""
 
         # Add the batch_size kwarg if not passed in and given a usable BATCH_SIZE
@@ -287,12 +297,13 @@ class EmbeddingModule(ModuleBase):
         """
         error.type_check("<NLP27491611E>", str, text=text)
 
-        embeddings = self._encode_with_retry(
+        embeddings, input_token_count = self._encode_with_retry(
             text, truncate_input_tokens=truncate_input_tokens
         )
         return EmbeddingResult(
             result=Vector1D.from_vector(embeddings),
             producer_id=self.PRODUCER_ID,
+            input_token_count=input_token_count,
         )
 
     @EmbeddingTasks.taskmethod()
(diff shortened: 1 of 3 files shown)


mynhardtburger:
FYI: @markstur

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -389,18 +402,21 @@ def run_sentence_similarities(
...
-        source_embedding = self._encode_with_retry(
+        source_embedding, input_token_count = self._encode_with_retry(
             source_sentences, truncate_input_tokens=truncate_input_tokens
         )
-        embeddings = self._encode_with_retry(
+        embeddings, input_token_count = self._encode_with_retry(
             sentences, truncate_input_tokens=truncate_input_tokens
         )
 
+        input_token_count = input_token_count + input_token_count
```
markstur, line 412: The same name was used, so 2nd overwrote the first.

mynhardtburger: fixed.

mynhardtburger:
_truncate_input_tokens()'s error case could also possibly be refactored to make use of _sum_token_count() to avoid having to rerun the tokenization.

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -352,17 +366,20 @@ def run_sentence_similarity(
             SentenceSimilarityResult: Similarity scores for each sentence.
         """
 
-        source_embedding = self._encode_with_retry(
+        source_embedding, source_token_count = self._encode_with_retry(
             source_sentence, truncate_input_tokens=truncate_input_tokens
         )
-        embeddings = self._encode_with_retry(
+        embeddings, embeddings_token_count = self._encode_with_retry(
```
markstur, line 372: nit: embeddings_token_count is a confusing var name. 

sentences_token_count or well even count2 would be better "embeddings_token_count", but doesn't matter. Varable name can be fixed whenever.

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -389,18 +406,21 @@ def run_sentence_similarities(
                 Each one contains the source-sentence's score for each sentence in order.
         """
 
-        source_embedding = self._encode_with_retry(
+        source_embedding, source_token_count = self._encode_with_retry(
             source_sentences, truncate_input_tokens=truncate_input_tokens
         )
-        embeddings = self._encode_with_retry(
+        embeddings, embeddings_token_count = self._encode_with_retry(
```
markstur, line 412: same here

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -670,10 +728,9 @@ def _truncate_input_tokens(
             okay_to_truncate = False
             max_length = max_tokens
 
-        if isinstance(texts[0], str):
-            to_tokenize = [texts]
-        else:
-            assert 0
+        to_tokenize = [texts]
```
markstur, line 731: given this change, I think this line can go after the asserts

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -728,18 +785,19 @@ def _truncate_input_tokens(
...
+        sentences: Union[str, Collection[str]],
         batch_size: int = 32,
-        device: str = None,
+        device: Optional[str] = None,
         convert_to_numpy: bool = True,
         convert_to_tensor: bool = False,
-        truncate_input_tokens: Optional[int] = 0,
-    ) -> np.ndarray:
+        truncate_input_tokens: int = 0,
+    ) -> EmbeddingResultTuple:
```
markstur, line 800: I guess it is necessary change the return type because we need more info than the array can give us.

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -728,18 +785,19 @@ def _truncate_input_tokens(
...
                 )
+        input_token_count = self._sum_token_count(tokenized)
 
-        return tokenized
+        return tokenized, input_token_count
 
     def encode(
         self,
-        sentences: Union[str, List[str]],
+        sentences: Union[str, Collection[str]],
```
markstur, line 794: We already have a signature does not match the base class situation (added an arg we need and chose to leave some out that we ignore), but I don't see much value in these new tweaks.  Were we already abusing the hints on those?

markstur (commented):
Need to bump minimum caikit version to get the proposed data model changes that are being used here.

Otherwise just some nits.

caikit_nlp/modules/text_embedding/embedding.py
mynhardtburger: It was an attempt to follow this [guidance](https://typing.readthedocs.io/en/latest/source/best_practices.html#arguments-and-return-types) which states to use abstract base classes for arguments and concrete types for return types.

I'm reverting this to `List[str]` since I also personally find it easier to read (the specifics of ABC's are not widely known).

caikit_nlp/modules/text_embedding/embedding.py
mynhardtburger: Renaming to `sentences_token_count`.

caikit_nlp/modules/text_embedding/embedding.py
mynhardtburger: Renaming to `sentences_token_count`.

caikit_nlp/modules/text_embedding/embedding.py
mynhardtburger: Agreed. reodering.

caikit_nlp/modules/text_embedding/embedding.py
mynhardtburger: My preference are to use named types for non-standard tuple returns. I find the development experience better.

evaline-ju:
@mynhardtburger caikit 0.26.14 is available with the data model update

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -639,9 +674,39 @@ def save(self, model_path: str, *args, **kwargs):
...
+        )
+        error.value_check(
+            "<NLP82314995E>",
+            tokenized.encodings,
+            "Number of tokenized encodings is only known when a non-python tokenizer is used",
+        )
+
+        token_count = 0
+
+        assert tokenized.encodings
```
evaline-ju, line 699: is this not covered by the `value_check`?

evaline-ju (commented):
Thanks for the contribution! caikit should now be ready to bump, and looks like linting might need some attention

caikit_nlp/modules/text_embedding/embedding.py
mynhardtburger: I'll remove.

Pylance can't figure out the type narrowing with `value_check()` (that tokenized.encodings can't be `None`). Just my habit of trying to eliminate red underlines.

tests/modules/text_embedding/test_embedding.py
```diff
@@ -788,3 +821,97 @@ def test_env_val_to_int():
...
+    [
+        ([0, 0, 0, 0, 0], [0]),
+        ([0, 1, 2, 3, 4], [0, 1, 2, 3, 4]),
+        ([0, 0, 1, 1, 1, 2], [0, 2, 5]),
+        ([], []),
+    ],
+)
+def test_get_sample_start_indexes(mapping, expected):
+    mock_tokenized = {"overflow_to_sample_mapping": mapping}
+    assert get_sample_start_indexes(mock_tokenized) == expected  # type: ignore
```
evaline-ju, line 917: q - are these not lists? (what is ignored here?)

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -767,26 +868,40 @@ def encode(
...
         ):  # Cast an individual sentence to a list with length 1
-            sentences = [sentences]
+            list_of_sentences = [sentences]
             input_was_string = True
 
+        error.type_check_all("<NLP82314994E>", str, sentences=list_of_sentences)
+
         if device is None:
-            device = self._target_device
+            device = self.device
```
evaline-ju, line 881: q - is this guaranteed to be set vs. `self._target_device`? Would this depend on another `.encode()` being called first (like the warmup)?

evaline-ju (commented):
A few questions, and it'd probably be good for @markstur to review again as well

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -842,11 +878,14 @@ def encode(
...
 
         self.to(device)
 
         all_embeddings = []
+
+        # Sort sentences according to length, from longest to shortest
+        # OOM errors then occurs at start of encoding
```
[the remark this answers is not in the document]
mynhardtburger: It is a comment I copied from the PR from `SentenceTransformer` where they explain why the sort order is used:
https://github.com/UKPLab/sentence-transformers/commit/ab7387c8956af840f55354f8981ba3fb2bfb6954

Since we are copying their code, I thought it was useful to at least partially explain why the sorting is done.

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -673,37 +674,84 @@ def save(self, model_path: str, *args, **kwargs):
...
-            token_count += sum(
-                (1 for t in tokenized.sequence_ids(batch_index=batch) if t is not None)
+def sum_token_count(
+    tokenized: BatchEncoding,
+    truncate_only: bool,
+) -> int:
+    """Returns the number of non-special tokens.
+    Args:
+        tokenized: BatchEncoding
+        only_truncated: bool
```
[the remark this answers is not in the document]
mynhardtburger: fixed

tests/modules/text_embedding/test_embedding.py
mynhardtburger: Thanks for this comment. This forced me to review and I found a bug!
I've updated the test inputs to be tensors. As a result I saw that `get_sample_start_indexes` also required a bug fix and the expected token count for `test_sum_token_count_with_truncation()` needed to be corrected. It previously counted all, instead of just the truncate component.

To answer your question:
- `mapping` is a is a `torch.Tensor` in a `BatchEncoding` object.
- It is difficult to directly or indirectly craft a `BatchEncoding` object precisely, hence why I opted to mock it. 
- `get_sample_start_indexes()` should return the index of the first occurance of each value in mapping:
`Tensor([0, 0, 1, 1, 1, 2])` -> `[0, 2, 5]` which represents the indexes in `Tensor([0, 0, 1, 1, 1, 2])` where the 1st occurrence of each number is located.

When input text is truncated by the tokenizer, it gets split into multiple encodings. `"overflow_to_sample_mapping"` holds the mapping of which encodings (index of mapping elements) relates to which input text (values of mapping elements).

tests/modules/text_embedding/test_embedding.py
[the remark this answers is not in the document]
mynhardtburger: I've added comments that the token counts are only valid for the `BertForSequenceClassification` model which is used for testing.

tests/modules/text_embedding/test_embedding.py
evaline-ju: thanks for the explanation! is the `# type: ignore` comment portion now removable then?

mynhardtburger: I'll remove the `#type: ignore`.
Linters will still complain since `get_sample_start_indexes` expects a `BatchEncoding` object while in the test it is provided a `dict` mock object with the minimum required keys to make `get_sample_start_indexes` work.

evaline-ju (approved):
LGTM!

markstur (approved):
LGTM.  Thanks!

PR #337, opened 2024-03-14
Make encode() in wrapped model compatible with super encode()

markstur:
* Adding missing params
* Don't return unexpected tuple (with token count) unless asked
* Adding check to not use our params if given an unwrapped model
* Fixing some param position things

### DIFF
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -66,13 +66,11 @@ try:
     sentence_transformers = importlib.import_module("sentence_transformers")
     # Third Party
     from sentence_transformers import SentenceTransformer
+    from sentence_transformers.util import batch_to_device, cos_sim, dot_score
     from sentence_transformers.util import (
-        batch_to_device,
-        cos_sim,
-        dot_score,
-        normalize_embeddings,
-        semantic_search,
+        normalize_embeddings as normalize,  # avoid parameter shadowing
     )
+    from sentence_transformers.util import semantic_search
 except ModuleNotFoundError:
     # When it is not available, create a dummy that raises an error on attempted init()
     class SentenceTransformerNotAvailable:
@@ -271,7 +269,9 @@ class EmbeddingModule(ModuleBase):
             exception=first_exception,
         )
 
-    def _encode_with_retry(self, *args, **kwargs) -> EmbeddingResultTuple:
+    def _encode_with_retry(
+        self, *args, **kwargs
+    ) -> Union[EmbeddingResultTuple, List[torch.Tensor], np.ndarray, torch.Tensor]:
         """All encode calls should use this for consistent param adding and retry loop"""
 
         # Add the batch_size kwarg if not passed in and given a usable BATCH_SIZE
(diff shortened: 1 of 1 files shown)


markstur:
@mynhardtburger

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -281,6 +281,16 @@ def _encode_with_retry(self, *args, **kwargs) -> EmbeddingResultTuple:
             if "batch_size" not in kwargs:
                 kwargs["batch_size"] = BATCH_SIZE
 
+        if isinstance(self.model, SentenceTransformerWithTruncate):
+            return self._with_retry(self.model.encode, *args, **kwargs)
+
+        # Else...
+        # It's possible to init with a model that doesn't have the added kwargs.
+        # E.g. a SentenceTransformer or other transormer model. Remove those kwargs!
```
evaline-ju, line 289:
```suggestion
        # E.g. a SentenceTransformer or other transformer model. Remove those kwargs!
```

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -837,32 +861,47 @@ def encode(
...
         :param truncate_input_tokens: Truncation length for input tokens.
                 Truncation length for input tokens.
                 If less than zero, this truncation is left up to the tokenizer default (model max).
                 If zero or greater than the model's maximum, then this is used as a test
                 to see if truncation is needed. If needed is needed, an exception is thrown.
                 Otherwise, we take this usable truncation limit to truncate the input tokens.
+        :param return_token_count: If true, a tuple is returned to add the input token count.
 
         :return:
            A tuple of the embedding, as a numpy matrix, and the input_token_count int.
```
evaline-ju, line 895: could this `return` be updated to reflect the `return_token_count` update?

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -281,6 +281,16 @@ def _encode_with_retry(self, *args, **kwargs) -> EmbeddingResultTuple:
...
+        if isinstance(self.model, SentenceTransformerWithTruncate):
+            return self._with_retry(self.model.encode, *args, **kwargs)
+
+        # Else...
+        # It's possible to init with a model that doesn't have the added kwargs.
+        # E.g. a SentenceTransformer or other transormer model. Remove those kwargs!
+        # This is not the normal use case but at least don't pass invalid kwargs, to encode()
+        # and don't return the unexpected tuple (adding token count).
+        del kwargs["truncate_input_tokens"]
+        del kwargs["return_token_count"]
```
evaline-ju, line 293: might be good to have a small test for this case, in case an update could break it (again)?

markstur: Yes and I came here to make the PR a draft because this isn't catching keyerrors (noticed while reviewing with Flavia)

caikit_nlp/modules/text_embedding/embedding.py
markstur: good catch, done

caikit_nlp/modules/text_embedding/embedding.py
markstur: almost missed this again, but done!

caikit_nlp/modules/text_embedding/embedding.py
markstur: Yep and done.  Definitely was a silly miss to not check AND not test. :(

evaline-ju (approved):
LGTM! Thanks for the test update!

PR #341, opened 2024-04-01
:zap: Tee stream instead of checking length

evaline-ju:
Functionality tested through `test_bootstrap_run_empty`


gkumbhat (approved):
Looks good to me! Thanks Evaline

PR #344, opened 2024-04-03
Embedding add a test that would have helped

markstur:
* Test for 2 identical strings to return identical vectors
* Also tests that batch vs single/loop are approx same
* This test would have found a bug that was recently fixed


evaline-ju (approved):
LGTM!

Issue #350, opened 2024-04-24
Tokenization with TGIS models resulting in bad wiretype


evaline-ju:
## Describe the bug

Tokenization tasks were added for modules that used TGIS in #325 in order to leverage the [`Tokenize` endpoint](https://github.com/IBM/text-generation-inference/blob/main/proto/generation.proto). However, when a GRPC `TokenizeTaskPredict` call was tried with a TGIS model, `grpc: failed to unmarshal the received message proto: bad wiretype` was seen. For REST, even when `/tokenization` was used, a generation result would be returned. It was observed that the TGIS `Generate` endpoint appeared to be used instead of the expected `Tokenize` one

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: Python 3.9
- Library version: Latest caikit-nlp 0.4.8

## Sample Code

`TokenizeTaskPredict` with a TGIS model as `mm-model-id` 

## Expected behavior

tokenization result

## Observed behavior

bad wiretype or generation result

PR #351, opened 2024-04-24
:sparkles: Add tokenization task to generation modules

evaline-ju:
Closes https://github.com/caikit/caikit-nlp/issues/350

Because `TextGenerationTGIS` is not a subclass but a backend type for `TextGeneration`, any additional tasks declared on the former were not actually getting added. By adding the additional tasks on `TextGeneration` and `PeftPromptTuning` respectively, this allows the tokenization tasks to be available on the TGIS backend implementations. Unimplemented functions have to be added or there will be errors.

Tokenization run functions could eventually be implemented since each LLM could be reasonably expected to have a tokenizer.

`Std` getting imported with `LaunchConfig` started to error in latest torch `2.3.0`. It was observed the `Std` object is at https://github.com/pytorch/pytorch/blame/main/torch/distributed/elastic/multiprocessing/api.py and not in the `LaunchConfig` and was potentially getting imported through another import to the launcher api. `tee` is no longer an arg on `LaunchConfig`. Since this was a breaking change, torch is now pinned below `2.3.0`.

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
+++ b/caikit_nlp/modules/text_generation/peft_prompt_tuning.py
@@ -48,8 +48,9 @@ from caikit.interfaces.nlp.data_model import (
     ClassificationTrainRecord,
     GeneratedTextResult,
     GeneratedTextStreamResult,
+    TokenizationResults,
 )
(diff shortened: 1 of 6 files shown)


gkumbhat (approved):
Looks good to me! Thanks for fixing.

PR #353, opened 2024-05-06
Expose model information for embeddings service

flaviabeo:
- Expose model information for embeddings service (`max_seq_lenght` +`sentence_embedding_dimension`) 
**Depends on**: https://github.com/caikit/caikit/pull/723

### DIFF
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -15,7 +15,7 @@
 # Standard
 from collections.abc import Sized
 from enum import Enum, auto
-from typing import Callable, Dict, List, NamedTuple, Optional, TypeVar, Union
+from typing import Any, Callable, Dict, List, NamedTuple, Optional, TypeVar, Union
 import importlib
 import os
 import time
@@ -177,6 +177,20 @@ class EmbeddingModule(ModuleBase):
         model.encode("warmup")
 
         return cls(model)
+
+    @property
+    def public_model_info(cls) -> Dict[str, Any]:  # pylint: disable=no-self-argument
+        """Helper property to return public metadata about a specific Model. This
+        function is separate from `metdata` as that contains the entire ModelConfig
+        which might not want to be shared/exposed.
+
+        Returns:
+            Dict[str, str]: A dictionary of this models's public metadata
+        """
+        return {
+            "max_seq_length": cls.model.max_seq_length,
+            "sentence_embedding_dimension": cls.model.get_sentence_embedding_dimension(),
+        }
 
     @classmethod
     def _get_ipex(cls, ipex_flag):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ classifiers=[
     "License :: OSI Approved :: Apache Software License"
 ]
 dependencies = [
-    "caikit[runtime-grpc,runtime-http]>=0.26.17,<0.27.0",
+    "caikit[runtime-grpc,runtime-http]>=0.26.27,<0.27.0",
     "caikit-tgis-backend>=0.1.27,<0.2.0",
     # TODO: loosen dependencies
     "accelerate>=0.22.0",
--- a/tests/modules/text_embedding/test_embedding.py
+++ b/tests/modules/text_embedding/test_embedding.py
@@ -197,6 +197,21 @@ def test_save_load_and_run():
     _assert_is_expected_embedding_result(result)
 
 
+def test_public_model_info():
+    """Check if we can get model info successfully"""
+    model_id = "model_id"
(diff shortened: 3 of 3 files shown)


caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -178,6 +180,22 @@ def load(cls, model_path: str, *args, **kwargs) -> "EmbeddingModule":
...
+
+        Returns:
+            Dict[str, str]: A dictionary of this models's public metadata
+        """
+        return {
+            "metadata": {
+                "max_seq_length": cls.model.max_seq_length,
+                "sentence_embedding_dimension": cls.model.get_sentence_embedding_dimension(),
+            }
+        }
```
gkumbhat, line 197: we can flatten it out and remove extra `metadata` key from here

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -661,6 +679,25 @@ def add_query(q):
...
+        """Run tokenization task against the embeddings model.
+
+        Args:
+           text: str
+                Text to tokenize
+        Returns:
+            TokenizationResults
+                The token count
+        """
+        if self.model:
```
gkumbhat, line 696: does every embeddings model supports tokenization via this method?

gkumbhat (changes requested):

caikit_nlp/modules/text_embedding/embedding.py
flaviabeo: I think every sentenceTransformer, yes -- as per the [docs in here](https://www.sbert.net/docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.tokenize) cc: @markstur

caikit_nlp/modules/text_embedding/embedding.py
gkumbhat: @flaviabeo , `cls` is conventionally used for `classmethods` (like `load`, `train` etc.).. However, since this is an instance method, we should be using `self` here instead of `cls`

flaviabeo: Hmm ok. I think there is an issue with that too, it is a property method from the base module, so when I use `self` instead of `cls` if prompts that I am changing the args from the base method -- like the error at the linting right now, using `self`, is:
```bash
caikit_nlp/modules/text_embedding/embedding.py:184:4: W0237: Parameter 'cls' has been renamed to 'self' in overriding 'EmbeddingModule.public_model_info' method (arguments-renamed)
```
Any idea how to go over this?

flaviabeo: And, when changed to `cls` this is the error:
```bash
caikit_nlp/modules/text_embedding/embedding.py:184:4: E0213: Method 'public_model_info' should have "self" as first argument (no-self-argument)
```

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -178,6 +180,20 @@ def load(cls, model_path: str, *args, **kwargs) -> "EmbeddingModule":
 
         return cls(model)
 
+    @property
+    def public_model_info(cls) -> Dict[str, Any]:
```
markstur, line 184: Does this help?

```suggestion
    def public_model_info(cls) -> Dict[str, Any]:  # pylint: disable=no-self-argument
```

flaviabeo: Thanks! It worked!!

caikit_nlp/modules/text_embedding/embedding.py
flaviabeo: @gkumbhat @evaline-ju issue with linting is solved!! May we move forward with this one?

flaviabeo: All checks have passed!

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -661,6 +677,25 @@ def add_query(q):
...
+        Args:
+           text: str
+                Text to tokenize
+        Returns:
+            TokenizationResults
+                The token count
+        """
+        if self.model:
+            return self.model.tokenize(
+                texts=text,
```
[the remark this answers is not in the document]
flaviabeo: The base method, if I am not mistaken, implements for string at [TokenizationTask(TaskBase)](https://github.com/caikit/caikit/blob/ed4c229ac4d7061d6f7ee06ce72bc0288eec6234/caikit/interfaces/nlp/tasks.py#L75), but I will run some tests and confirm the behavior. Thank you!

gkumbhat (approved):
Looks good to me!

Issue #354, opened 2024-05-13
Do not surface `grpc.RpcError` from downstream calls to TGIS


gabe-l-hart:
## Describe the bug

The modules that call downstream to TGIS need to have consistent semantics with other local modules for errors. Currently, the [tgis_utils](https://github.com/caikit/caikit-nlp/blob/main/caikit_nlp/toolkit/text_generation/tgis_utils.py) utilities do not catch errors in the downstream calls, and neither do the `modules` that call them, so these errors are percolating up to the global `try/except` in the `caikit.runtime` server(s) when these modules are served.

When using the `grpc_server`, the global `try/except` already handles `RpcError` since it is a `grpc` server, but the `http_server` does not which results in a 500-class error to the user, even when the underlying TGIS exception is 400-class (e.g. `INVALID_ARGUMENT`). Since the `http_server` is generic, it should not be responsible for knowing about how the modules in `caikit_nlp` raise exceptions and should instead be able to expect that all exceptions are raised as either `CaikitCoreException` or `CaikitRuntimeException`.

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: any
- Library version: `0.4.9`

## Sample Code

```py
import grpc
import caikit_tgis_backend
import caikit
from caikit_nlp.toolkit.text_generation.tgis_utils import TGISGenerationClient
from caikit.core.exceptions.caikit_core_exception import CaikitCoreException
from caikit.runtime.types.caikit_runtime_exception import CaikitRuntimeException
from concurrent.futures import ThreadPoolExecutor


class BadTgis(caikit_tgis_backend.protobufs.generation_pb2_grpc.GenerationServiceServicer):
    def Generate(self, request, context):
        context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
        context.set_details("Yikes!")
        return
    def GenerateStream(self, request, context):
        context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
        context.set_details("Streaming Yikes!")


# Build the server
svc = BadTgis()
pool = ThreadPoolExecutor(max_workers=1)
server = grpc.server(thread_pool=pool)
caikit_tgis_backend.protobufs.generation_pb2_grpc.add_GenerationServiceServicer_to_server(svc, server)
server.add_insecure_port("[::]:12345")
server.start()

# Create the utils client
[...]

PR #355, opened 2024-05-13
HandleRpcError: Add RpcError handling in tgis_utils

gabe-l-hart:
## Description

This PR adds error handling for `grpc.RpcError`s raised during calls to TGIS in `tgis_utils`.

Closes: https://github.com/caikit/caikit-nlp/issues/354


caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -41,20 +48,53 @@
...
+    grpc.StatusCode.UNKNOWN: CaikitCoreStatusCode.UNKNOWN,
+    grpc.StatusCode.INVALID_ARGUMENT: CaikitCoreStatusCode.INVALID_ARGUMENT,
+    grpc.StatusCode.DEADLINE_EXCEEDED: CaikitCoreStatusCode.CONNECTION_ERROR,
+    grpc.StatusCode.NOT_FOUND: CaikitCoreStatusCode.NOT_FOUND,
+    grpc.StatusCode.ALREADY_EXISTS: CaikitCoreStatusCode.INVALID_ARGUMENT,
+    grpc.StatusCode.PERMISSION_DENIED: CaikitCoreStatusCode.FORBIDDEN,
+    grpc.StatusCode.RESOURCE_EXHAUSTED: CaikitCoreStatusCode.UNKNOWN,
+    grpc.StatusCode.FAILED_PRECONDITION: CaikitCoreStatusCode.UNKNOWN,
+    grpc.StatusCode.ABORTED: CaikitCoreStatusCode.CONNECTION_ERROR,
+    grpc.StatusCode.OUT_OF_RANGE: CaikitCoreStatusCode.UNKNOWN,
```
gkumbhat, line 78: Conversions are lossy, but kind of make sense from changing the view of runtime to library.... For some of these, like the `OUT_OF_RANGE`, might make sense to expose at library level as well ?  Or may be we can map it to `FAILED_PRE_CONDITION` since there is a overlap as per doc [here](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)

Also the `FAILED_PRECONDITION`  might come in handy, which is getting mapped to `UNKNOWN` above 🤔

gkumbhat (commented):
Left a small mapping suggestion / question. Otherwise looks good to me!

caikit_nlp/toolkit/text_generation/tgis_utils.py
gabe-l-hart: Yeah, I was trying to make this work without changing `caikit` itself. I think there is almost certainly an important API-breaking change conversation to be had about how the internal error codes map to well-known error code enums (`grpc`/`http`) and how to avoid information loss.

gkumbhat (approved):
LGTM. We do require some more iteration w.r.t the error codes, but some changes need to be made on `caikit` side to include those codes. So for now, we can get this in and iterate on error codes..

However, iterating on error codes would be API breaking. :rotating_light:

gkumbhat (approved):
LGTM.  Per the discussion, we will modify and add error code mapping in caikit itself in future, which would allow us to remove the mapping from caikit-nlp

PR #356, opened 2024-05-14
prevent pip backtracking

dtrifiro:
image builds are currently taking 5+h and failing because of pip backtracking.

- deps: pin grpc-\* to a recent version to avoid pip backtracking
- Dockerfile: create venv with updated pip/setuptools

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,9 @@ dependencies = [
     "caikit[runtime-grpc,runtime-http]>=0.26.17,<0.27.0",
     "caikit-tgis-backend>=0.1.27,<0.2.0",
     # TODO: loosen dependencies
+    "grpcio>=1.62.2", # explicitly pin grpc dependencies to a recent version to avoid pip backtracking
+    "grpcio-reflection>=1.62.2",
+    "grpcio-health-checking>=1.62.2",
     "accelerate>=0.22.0",
     "datasets>=2.4.0",
     "huggingface-hub",
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN --mount=source=.git,target=.git,type=bind tox -e build
 
 FROM base as deploy
 
-RUN python -m venv /opt/caikit/
+RUN python -m venv --upgrade-deps /opt/caikit/
 
 ENV VIRTUAL_ENV=/opt/caikit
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -17,6 +17,9 @@ dependencies = [
     "caikit[runtime-grpc,runtime-http]>=0.26.17,<0.27.0",
     "caikit-tgis-backend>=0.1.27,<0.2.0",
     # TODO: loosen dependencies
+    "grpcio>=1.62.2", # explicitly pin grpc dependencies to a recent version to avoid pip backtracking
+    "grpcio-reflection>=1.62.2",
+    "grpcio-health-checking>=1.62.2",
```
evaline-ju, line 22: These are technically `caikit`/`caikit[runtime-grpc]` dependencies, and we tended to keep a fairly loose versioning system on caikit to allow any applications/users potentially specify tighter bounds. 

I'm fine with this (forcing users on much more recent versions) in theory here since this library is likely not getting very wide use, but I am curious - how is this particular `Dockerfile`/ image build currently being used? I presume https://github.com/opendatahub-io/caikit-nlp is the one building https://quay.io/repository/opendatahub/caikit-nlp?tab=tags ?

gkumbhat: it would be ideal if they are in pinned in `caikit/ caikit[runtime-grpc]` , that way other dependent library also get it. Since grpc is internal to `caikit-runtime` it kind of becomes confusing to add them here. But these can be updated here as an "override" for now though.

dtrifiro: We could pin those dependencies in `caikit` instead if you'd prefer that, I preferred pinning them here because this is where the issue was encountered first, although I'm not sure if the backtracking is caused by `caikit[runtime-grpc]` or itself some other dependency added by caikit-nlp. I can investigate if you'd prefer.

Regarding the `Dockerfile`, that's correct, the image is the one published at `quay.io/opendatahub/caikit-nlp`

gkumbhat: yeah, the gRPC dependency is primarily coming from `runtime`, but we can unblock the build here quickly by setting these overrides here. It may be possible that `caikit` doesn't see this resolution slowdown, because it doesn't have conflicting deps, so it may as well need to be here as overrides!

gkumbhat (approved):
Looks good to me!

PR #359, opened 2024-05-17
Add graph mode to embedding models which benefits from ipex optimize

devpramod:
Graph mode (torchscript) is an additional step that can further accelerate a workload that has been optimized with ipex and bfloat16 along with mixed precision. Graph mode utilizes AMX on SPR for additional speedups.

This PR modifies config file to have `graphmode` as an option to end users
The graph is compiled in the constructor of the embedding module such that it is ready for execution when requests arrive.

### DIFF
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -916,8 +916,8 @@ class SentenceTransformerWithTruncate(SentenceTransformer):
 
     def _apply_graph_mode(self) -> torch.jit.ScriptModule:
         """
-        Compiles the model into a TorchScript graph using predefined fixed-size randomized 
-        input tensors.The tensors simulate typical input structures without relying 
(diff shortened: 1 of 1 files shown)


gkumbhat (changes requested):
Thanks for the contribution @devpramod ... 

Couple of tiny housekeeping related comments for the PR:
1. Can you rebase the PR with a `sign-off`, this allows github to verify and the DCO check would pass.
2. It seems the lint and formatting checks are failing. Can you please address those.

Guidelines for contribution documenting some of these steps can be found here: https://github.com/caikit/caikit-nlp/blob/main/CONTRIBUTING.md


On the graphmode integration, couple of question and suggestions:
1. When should the graph model be enabled? Can you add describe a bit when it is applicable, for which models etc ?
2.  How much of speedup is expected with enablement of `graphmode` ? Can you share some results ?

devpramod:
Hi @gkumbhat 
I have resolved the formatting, linting and DCO check issues

PR #363, opened 2024-06-06
Lazy model connection & x-route-info

mynhardtburger:
This PR adds support for respecting a TGIS hostname override (`x-route-info` header/metadata).

This is achieved by lazily registering the model connection with the TGISBackend when `.run()`/`run_stream_out()`/`run_tokenizer()` is called. The `x-route-info` header/metadata is read from the context arg.

Additionally:
- a `TextGenerationTGIS._tgis_backend` property was added to be consistent with `PeftPromptTuningTGIS`. The `TextGenerationTGIS.tgis_backend` property is conditional on the `tgis_backend` argument to be True, which made accessing `TextGenerationTGIS.tgis_backend` risky as it might not exist. The additional `_tgis_backend` property is guaranteed to exist, making using it simpler and less risky.

Depends on `caikit-tgis-backend>=0.1.33`

### DIFF
--- a/caikit_nlp/modules/text_generation/peft_tgis_remote.py
+++ b/caikit_nlp/modules/text_generation/peft_tgis_remote.py
@@ -73,7 +73,6 @@ class PeftPromptTuningTGIS(ModuleBase):  # pylint: disable=too-many-instance-att
     ) -> None:
         super().__init__()
 
-        # self._client = None
         self._tgis_backend = tgis_backend
         if enable_backend:
             error.type_check(
--- a/caikit_nlp/modules/text_generation/peft_tgis_remote.py
+++ b/caikit_nlp/modules/text_generation/peft_tgis_remote.py
@@ -233,8 +233,7 @@ class PeftPromptTuningTGIS(ModuleBase):  # pylint: disable=too-many-instance-att
             self.enable_backend,
             "Backend must be configured and loaded with this module before executing `run` call.",
         )
-        if self._tgis_backend:
-            self._register_model_connection_with_context(context)
+        self._register_model_connection_with_context(context)
 
         verbalized_text = render_verbalizer(self.verbalizer, {"input": text})
         return self.tgis_generation_client.unary_generate(
@@ -300,8 +299,7 @@ class PeftPromptTuningTGIS(ModuleBase):  # pylint: disable=too-many-instance-att
             before executing `run_stream_out` call.",
         )
 
-        if self._tgis_backend:
-            self._register_model_connection_with_context(context)
+        self._register_model_connection_with_context(context)
 
         verbalized_text = render_verbalizer(self.verbalizer, {"input": text})
         return self.tgis_generation_client.stream_generate(
@@ -341,8 +339,8 @@ class PeftPromptTuningTGIS(ModuleBase):  # pylint: disable=too-many-instance-att
             TokenizationResults
                 The token count
         """
-        if self._tgis_backend:
-            self._register_model_connection_with_context(context)
+
+        self._register_model_connection_with_context(context)
 
         return self.tgis_generation_client.unary_tokenize(
             text=text,
@@ -352,18 +350,19 @@ class PeftPromptTuningTGIS(ModuleBase):  # pylint: disable=too-many-instance-att
         self, context: Optional[RuntimeServerContextType]
     ):
         """
-        Register a model connection with the configured TGISBackend.
+        Register a remote model connection with the configured TGISBackend if there is
+        a context override provided.
         """
-        ok, route_info = get_route_info(context)
-        if ok:
-            log.debug(
-                "<NLP10705560D> Registering remote model connection with context "
-                "override: 'hostname: %s'",
-                route_info,
-            )
-            self._tgis_backend.register_model_connection(
-                self.base_model_name, {"hostname": route_info}, fill_with_defaults=True
-            )
-        else:
-            self._tgis_backend.register_model_connection(self.base_model_name)
-        self._model_loaded = True
+        if self._tgis_backend:
+            if route_info := get_route_info(context):
+                log.debug(
+                    "<NLP10705560D> Registering remote model connection with context "
+                    "override: 'hostname: %s'",
+                    route_info,
+                )
+                self._tgis_backend.register_model_connection(
+                    self.base_model_name,
+                    {"hostname": route_info},
+                    fill_with_defaults=True,
+                )
+            self._model_loaded = True
--- a/caikit_nlp/modules/text_generation/peft_tgis_remote.py
+++ b/caikit_nlp/modules/text_generation/peft_tgis_remote.py
@@ -14,7 +14,9 @@
 """This file contains a distributed backend implementation for leveraging the PEFT-trained
 prompt vectors in TGIS generation requests.
 """
+
 # Standard
+from functools import cached_property
 from typing import Iterable, List, Optional, Tuple, Union
 import os
 
@@ -32,6 +34,7 @@ from caikit.interfaces.nlp.data_model import (
     TokenizationResults,
 )
 from caikit.interfaces.nlp.tasks import TextGenerationTask, TokenizationTask
+from caikit.interfaces.runtime.data_model import RuntimeServerContextType
 from caikit_tgis_backend import TGISBackend
 import alog
 
@@ -40,6 +43,7 @@ from ...data_model import ExponentialDecayLengthPenalty
 from ...toolkit.text_generation.tgis_utils import (
     GENERATE_FUNCTION_TGIS_ARGS,
     TGISGenerationClient,
+    get_route_info,
 )
 from ...toolkit.verbalizer_utils import render_verbalizer
 from . import PeftPromptTuning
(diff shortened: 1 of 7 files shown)


caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -107,6 +110,14 @@ def __del__(self):
             if tgis_backend and prompt_cache_id and model_id:
                 tgis_backend.unload_prompt_artifacts(model_id, prompt_cache_id)
 
+    @cached_property
+    def _client(self):
+        # Configure the internal client
+        # NOTE: This is made optional for the cases where we do not need to execute `.run` function
+        # for example, bootstrapping a model to caikit format and saving.
+        if hasattr(self, "tgis_backend") and self._tgis_backend:
+            return self._tgis_backend.get_client(self.base_model_name)
```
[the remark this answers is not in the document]
mynhardtburger: Added a `_tgis_backend` property which is guaranteed to exist. Allows us to do a simple `if self._tgis_backend` instead of using `getattr()` which breaks intellisense/autocomplete in IDEs.

caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -320,6 +341,21 @@ def run_tokenizer(
...
             text=text,
         )
+
+    def _register_model_connection_with_context(
+        self, context: Optional[RuntimeServerContextType]
+    ):
+        ok, route_info = get_route_info(context)
+        if ok:
+            self._tgis_backend.register_model_connection(
+                self.base_model_name, {"hostname": route_info}
```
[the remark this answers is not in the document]
mynhardtburger: The `fill_with_defaults` arg has been added and used.

caikit_nlp/modules/text_generation/text_generation_tgis.py
```diff
@@ -109,6 +105,14 @@ def __del__(self):
         if self._model_loaded and self.tgis_backend:
             self.tgis_backend.unload_model(self.model_name)
 
+    @cached_property
+    def _client(self):
+        # Configure the internal client
+        # NOTE: This is made optional for the cases where we do not need to execute `.run` function
+        # for example, bootstrapping a model to caikit format and saving.
+        if hasattr(self, "tgis_backend") and self.tgis_backend:
+            return self.tgis_backend.get_client(self.model_name)
```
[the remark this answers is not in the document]
mynhardtburger: Added a `_tgis_backend` property which is guaranteed to exist. Allows us to do a simple `if self._tgis_backend` instead of using `getattr()` which breaks intellisense/autocomplete in IDEs.

caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -683,3 +685,31 @@ def unary_tokenize(
...
+    """
+    Returns a tuple `(True, x-route-info)` from context if "x-route-info" was found in the headers/metadata.
+
+    Otherwise returns a tuple `(False, None)` if "x-route-info" was not found in the context or if context is None.
+    """
+    if context is None:
+        return False, None
+
+    if isinstance(context, grpc.ServicerContext):
+        route_info = dict(context.invocation_metadata()).get("x-route-info")
```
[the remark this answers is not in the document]
mynhardtburger: Done. Moved to a global constant.

caikit_nlp/modules/text_generation/text_generation_tgis.py
```diff
@@ -333,7 +345,22 @@ def run_tokenizer(
...
                 text=text,
             )
+
+    def _register_model_connection_with_context(
+        self, context: Optional[RuntimeServerContextType]
+    ):
+        ok, route_info = get_route_info(context)
+        if ok:
+            self.tgis_backend.register_model_connection(
+                self.model_name, {"hostname": route_info}
```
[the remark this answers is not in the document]
mynhardtburger: The `fill_with_defaults` arg has been added and used.

caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -221,6 +233,9 @@ def run(
             self.enable_backend,
             "Backend must be configured and loaded with this module before executing `run` call.",
         )
+        if self._tgis_backend:
```
gabe-l-hart, line 236: Total NIT, but since `_register_model_connection_with_context` is a member function, you could move the `if self._tgis_backend` line into the function and save a few duplicate lines in each of the inference methods.

caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -320,6 +341,28 @@ def run_tokenizer(
...
         )
+
+    def _register_model_connection_with_context(
+        self, context: Optional[RuntimeServerContextType]
+    ):
+        """
+        Register a model connection with the configured TGISBackend.
+        """
+        ok, route_info = get_route_info(context)
+        if ok:
```
gabe-l-hart, line 358: I really want to be able to use `:=` here and make it look more like `go`, but seems that [you can't](https://stackoverflow.com/a/59568122)

caikit_nlp/modules/text_generation/peft_tgis_remote.py
```diff
@@ -320,6 +341,28 @@ def run_tokenizer(
...
+            log.debug(
+                "<NLP10705560D> Registering remote model connection with context override: 'hostname: %s'",
+                route_info,
+            )
+            self._tgis_backend.register_model_connection(
+                self.base_model_name, {"hostname": route_info}, fill_with_defaults=True
+            )
+        else:
+            self._tgis_backend.register_model_connection(self.base_model_name)
```
gabe-l-hart, line 368: I think we don't need to do this if there's no context because it will be done automatically with `get_client`

caikit_nlp/modules/text_generation/text_generation_tgis.py
```diff
@@ -240,6 +249,9 @@ def run(
             GeneratedTextResult
                 Generated text result produced by TGIS.
         """
+        if self._tgis_backend:
```
gabe-l-hart, line 252: Same comment about burying the `if self._tgis_backend`

caikit_nlp/modules/text_generation/text_generation_tgis.py
```diff
@@ -333,7 +349,26 @@ def run_tokenizer(
...
+        if ok:
+            log.debug(
+                "<NLP15770311D> Registering remote model connection with context override: 'hostname: %s'",
+                route_info,
+            )
+            self._tgis_backend.register_model_connection(
+                self.model_name, {"hostname": route_info}, fill_with_defaults=True
+            )
+        else:
+            self._tgis_backend.register_model_connection(self.model_name)
```
gabe-l-hart, line 374: Same comment about not needing the `else` clause

caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -84,6 +86,8 @@
     grpc.StatusCode.UNAUTHENTICATED: CaikitCoreStatusCode.UNAUTHORIZED,
 }
 
+ROUTE_INFO_KEY = "x-route-info"
```
gabe-l-hart, line 89: NIT:
```suggestion
# HTTP Header / gRPC Metadata key used to identify a route override
ROUTE_INFO_HEADER_KEY = "x-route-info"
```

caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -683,3 +687,33 @@ def unary_tokenize(
...
+def get_route_info(
+    context: Optional[RuntimeServerContextType],
+) -> Tuple[bool, Optional[str]]:
+    """
+    Returns a tuple `(True, x-route-info)` from context if "x-route-info" was found in
+    the headers/metadata.
+
+    Otherwise returns a tuple `(False, None)` if "x-route-info" was not found in the
+    context or if context is None.
+    """
```
gabe-l-hart, line 701: I think we can collapse the return into a simple `Optional[str]`. Returning `None` is indicative of "no route info found." This would simplify the logic that uses it and allow us to get another walrus!

tests/fixtures/__init__.py
```diff
@@ -342,3 +343,15 @@ def temp_config(**overrides):
 
     with mock.patch.object(caikit.config.config, "_IMMUTABLE_CONFIG", local_config):
         yield local_config
+
+
+class TestServicerContext:
+    """
+    A dummy class for mimicking ServicerContext invocation metadata storage.
```
gabe-l-hart, line 350: Interesting. I was looking for a link to suggest as a citation of this interface and I found [this](https://github.com/grpc/grpc/blob/d038e0ed1970d8808e0f1e81acbefeb2f75f5941/src/python/grpcio_testing/grpc_testing/_server/_servicer_context.py#L19) instead. Makes me wonder if we should be using [grpcio-testing](https://pypi.org/project/grpcio-testing/) for all of this in our tests!

gabe-l-hart: Here's the citation link: https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/__init__.py#L1128

gabe-l-hart (changes requested):
A couple of small cleanup suggestions and comment requests

caikit_nlp/toolkit/text_generation/tgis_utils.py
```diff
@@ -683,3 +688,33 @@ def unary_tokenize(
...
+        route_info = context.headers.get(ROUTE_INFO_HEADER_KEY)
+        if route_info:
+            return route_info
+    else:
+        error.log_raise(
+            "<NLP92615097E>",
+            ValueError(f"context is of an unsupported type: {type(context)}"),
+        )
+
+    return None
```
gabe-l-hart, line 720: NIT: this is both implicit and unreachable since `else` will raise. I'm guessing it's here because your IDE doesn't know that?

gabe-l-hart (approved):
One list NIT, but I think this is also good as is. Thanks!

caikit_nlp/toolkit/text_generation/tgis_utils.py
mynhardtburger: Done. Removed.

caikit_nlp/toolkit/text_generation/tgis_utils.py
mynhardtburger: Done. Now using walrus operator for good @gabe-l-hart  karma.

caikit_nlp/modules/text_generation/text_generation_tgis.py
mynhardtburger: done.

PR #366, opened 2024-07-03
Fix CVE-2024-5480

vaibhavjainwiz:
Update floor value of torch version to 2.2.2 so fix CVE
https://nvd.nist.gov/vuln/detail/CVE-2024-5480


gkumbhat (approved):
Looks good to me!

PR #367, opened 2024-07-09
Add param include_stop_sequence

waleedqk:
The `include_stop_sequence` option is provided by tgis, which the user can set to include the STOP_SEQUENCE in the output.
This PR makes the option available in the FastAPI endpoint


gkumbhat (approved):
Looks good to me!

PR #369, opened 2024-07-22
Fix embedding use of tokenizer to avoid already borrowed

markstur:
With high concurrency the fast tokenizer (Rust) hits "Already borrowed" exceptions with the truncation and padding (and max length) params change. Do not change those. Detect where to truncate (or return message) using the default settings and some code. Truncate texts and re-tokenize when needed.

In addition, the error message which used to report number of tokens over the limit (but did not say which sentence) now returns the index(es) of the sentence(s) that was/were found to be too long.


caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -925,11 +956,22 @@ def _tokenize_plus(
...
+        Use a copy of the tokenizer per-model (self) and per-thread (map by thread ID).
+
         Avoid changing the max length, truncation, and padding to avoid the
         "Already borrowed" errors that come with concurrent threads attempting to use
         the fast tokenizer with different truncation settings.
         """
-        return self.tokenizer(
+
+        # Keep copies of tokenizer per thread (in each wrapped model instance)
+        thread_id = threading.get_ident()
```
gabe-l-hart, line 967: One other primitive to take a look at here is [threading.Local](https://docs.python.org/3/library/threading.html#thread-local-data)

markstur: I tried using threading.Local instead of a map, but realized that each thread would need a map to get the correct tokenizer per model.  So I preferred map keyed by thread in the model instance vs map keyed by model name (id?) in thread local.   So I gave up experimenting with threading.Local, but was hoping it had some advantage here that I did not get to.  Is a map in thread local better somehow for this?

gabe-l-hart: Ah, makes sense!

PR #373, opened 2024-08-05
:arrow_up: Bump upper bound of torch

evaline-ju:
Include latest 2.4 version of torch


gkumbhat (commented):
Looks like this one requires rebase

PR #375, opened 2024-08-05
:construction_worker: Reclaim space on image build

evaline-ju:
We've seen multiple PR build failures on the image build workflow with `buildx failed with: ERROR: failed to solve: rpc error: code = Unknown desc = write /opt/caikit/lib/python3.9/site-packages/torch/lib/libtorch_cpu.so: no space left on device`


gkumbhat (approved):
Thanks Evaline. This gets auto-tested with all checks passing on the PR.

PR #384, opened 2024-08-16
:goal_net: Handle error details on inactive rpc errors

evaline-ju:
Previously #372 was added to give connection errors a more understandable user message, but this obscured some unary generation validation errors

In this PR:
- Serializing `err.details` was leading to JSON errors - this should be `err.details()`
- Non-connection errors such as validation errors [invalid argument] were getting surfaced with the inactive RPC connection message incorrectly

Test command with this fix
```
grpcurl -plaintext -H "mm-model-id: [model-id]" -d '{
  "stopSequences": [],
  "text": "sad",
  "tokenLogprobs": true
}' localhost:8085 caikit.runtime.Nlp.NlpService.TextGenerationTaskPredict
```


gkumbhat (approved):
🤦  thanks for fixing it. LGTM.

PR #393, opened 2024-10-08
Enable using kwargs for selecting pad-to-max-length strategy for tokenizer in embeddings

kcirred:
Allows users to pass tokenizer keyword argument to select the tokenizer settings desired. In this case, enabling for pad_to_max_length.

### DIFF
--- a/caikit_nlp/modules/text_embedding/embedding.py
+++ b/caikit_nlp/modules/text_embedding/embedding.py
@@ -37,7 +37,6 @@ import time
 from torch import nn
 from torch.backends import mps
 from transformers import BatchEncoding
-from transformers.tokenization_utils import PaddingStrategy
 import numpy as np
 import torch
 
@@ -1041,7 +1040,7 @@ class SentenceTransformerWithTruncate(SentenceTransformer):
         the fast tokenizer with different truncation settings.
         """
 
-        pad_to_max_length = kwargs.pop("pad_to_max_length", None)
+        padding_strategy = kwargs.pop("padding_strategy", True)
 
         # Keep copies of tokenizer per thread (in each wrapped model instance)
         thread_id = threading.get_ident()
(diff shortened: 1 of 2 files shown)


caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -1047,18 +1051,32 @@ def _get_tokenized(self, texts):
...
-            return_overflowing_tokens=False,  # DO NOT USE overflow tokens break sentence batches
-            return_offsets_mapping=True,  # Used for truncation
-            return_length=False,
-            return_tensors="pt",
-            truncation=True,  # DO NOT CHANGE else "Already borrowed" errors
-            padding=True,  # DO NOT CHANGE else "Already borrowed" errors
-            max_length=self.max_seq_length,  # DO NOT CHANGE else "Already borrowed" errors
-        )
+        if pad_to_max_length:
+                return tokenizer(
```
gabe-l-hart, line 1055: I think indentation got funky on this line

gabe-l-hart (changes requested):
Thanks for the contribution! I'm going to defer full review to @gkumbhat, but there's one little formatting NIT I noticed when glancing through it.

caikit_nlp/modules/text_embedding/embedding.py
```diff
@@ -1039,6 +1041,8 @@ def _get_tokenized(self, texts):
         the fast tokenizer with different truncation settings.
         """
 
+        pad_to_max_length = kwargs.pop("pad_to_max_length", None)
```
gkumbhat, line 1044: Question: is there a reason to accept `pad_to_max_length` as value here instead of just `max_length` which can then be directly passed to `tokenizer` function ? 

https://huggingface.co/docs/transformers/en/main_classes/tokenizer#transformers.PreTrainedTokenizer.__call__.padding

gkumbhat (changes requested):
Thanks for the contribution.

Looks like DCO is failing on this PR. Can you do a `--signoff` and push again?

kcirred:
@gkumbhat I also wrote a unit test so let me push that as well. 

I suspect the `def sum_token_count` in `embeddings.py` is doing what is intended. By counting `sum(encoding.attention_mask)` the 0s will be dropped. 

In my test case, I originally wanted to show that the results stays the same but there will be a change to the tokenizer `input_ids` and `attention_mask` shape. I will not use `input_token_count` which comes from `sum_token_count` as they will come out to be the same despite the change in `attention_mask`. For now I only test that the change in tokenizer option in our use case does not change the final result.

caikit_nlp/modules/text_embedding/embedding.py
kcirred: @gkumbhat we can change the variable name based on your suggestion. The goal is to make sure that this variable will set the tokenizer to pad to max sequence length of the model.

gkumbhat (approved):
Looks good to me! Thanks for making updates

gkumbhat (approved):
Looks good to me!

gabe-l-hart (approved):
Looks good!

PR #400, opened 2024-10-28
[embeddings] extend kwargs to functions that call _encode_with_retry

kcirred:


gkumbhat (commented):
Hi @kcirred , can you please provide a bit of context behind why this PR is required, what it does, how can user use this change and how was it tested ?

gkumbhat (approved):
Looks good to me!

The details of the PR were discussed separately!

PR #413, opened 2025-02-24
:ambulance: Fix tf keras package creating confusion at install time

gkumbhat:
### Changes

- Move `tf-keras` to test dependencies as they are only required for working with the dummy models at test time.


evaline-ju (approved):
LGTM!

PR #414, opened 2025-02-28
Update verbalizer error message to include the relevant template string

mynhardtburger:
Update the exception message to include the template field which is the cause for the error.

### DIFF
--- a/caikit_nlp/toolkit/verbalizer_utils.py
+++ b/caikit_nlp/toolkit/verbalizer_utils.py
@@ -61,7 +61,7 @@ def render_verbalizer(verbalizer_template: str, source_object) -> str:
                {{output}} with getattr(source_object, "target").
             returns: "Source: machine Target: learning"
 
-    NOTE: This function will throw  KeyError/AttributeError if you try to grab a key or property
+    NOTE: This function will throw ValueError if you try to grab a key or property
     that is invalid.
 
     Args:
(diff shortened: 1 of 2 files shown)


caikit_nlp/toolkit/verbalizer_utils.py
```diff
@@ -89,15 +89,19 @@ def replace_text(match_obj):
             if index_object not in source_object:
                 error(
                     "<NLP97415192E>",
-                    KeyError("Requested template string is not a valid key in dict"),
+                    KeyError(
+                        "Requested template string '%s' is not a valid key in dict",
+                        index_object,
+                    ),
```
gkumbhat: nit: can we change it to inline f string format:

```suggestion
                    KeyError(
                        f"Requested template string '{index_object}' is not a valid key in dict"
                    ),
```

caikit_nlp/toolkit/verbalizer_utils.py
```diff
@@ -89,15 +89,19 @@ def replace_text(match_obj):
...
                 )
             return source_object[index_object]
 
         if not hasattr(source_object, index_object):
             error(
                 "<NLP97715112E>",
                 AttributeError(
-                    "Requested template string is not a valid property of type"
+                    "Requested template string '%s' is not a valid property of type",
+                    index_object,
```
gkumbhat: nit: same here
```suggestion
                    f"Requested template string '{index_object}' is not a valid property of type",
```

gkumbhat: also we should check if this `index_object` is actually a string, otherwise might cause problem without explicit stringification

mynhardtburger: I don't think it is possible for the fstring to fail no matter the type of `index_object`, unless the `__str__` dunder method was overwritten and that raises an exception.

By default `str()` is used to serialize the object. At worst the the output might be in `repr()` format.

gkumbhat (approved):
Looks good to me!

PR #429, opened 2025-07-21
:arrow_up: Bump caikit range

evaline-ju:
- Allow more flexible caikit versioning including dependent protobuf versions
- numpy>=2 has not been tested with caikit-nlp, so this is pinned


gkumbhat (approved):
Looks good to me!

PR #439, opened 2025-11-18
Fix CVE's for caikit-nlp

rpancham:
Addresses: https://issues.redhat.com/browse/RHOAIENG-37431


RH-steve-grubb (approved):

edwardquarm (approved):
lgtm

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "scipy>=1.10.0",
     "sentence-transformers>=3.4.0,<3.5.0",
     "tokenizers>=0.20.0",
-    "torch>=2.3.1,<2.9.0",
+    "torch>=2.8.0",
```
gabe-l-hart, line 32: In general, we try to avoid _requiring_ a lower bound like this since it disallows users that are stuck on older versions for one reason or another. I think the correct thing to do here is to simply remove the upper bound:

```suggestion
    "torch>=2.3.1",
```

gabe-l-hart (changes requested):
We'd like to keep support for older torch versions possible while also removing the upper bound. Assuming tests pass with the latest `torch`, this should be merge ready once we restore the known-working lower bound of `2.3.1`.

pyproject.toml
alex-jw-brooks: @rpancham @gabe-l-hart is this change actually needed, or do we just need to explicitly build an image somewhere with torch 2.8?

I followed the CVEs, and it seems to be patched in 2.8+, but that should already be compatible with things as they are, right?
https://github.com/advisories/GHSA-887c-mr87-cxwp
https://github.com/advisories/GHSA-3749-ghw9-m3mg

rpancham: As per my understanding, the CVE fix for PyTorch (https://github.com/advisories/GHSA-887c-mr87-cxwp) is included in torch 2.8.0+, so ensuring our image/tests use a patched version is sufficient from a security perspective. Because of that, we don’t need to keep the upper bound in pyproject.toml.

If the goal is to maintain compatibility with older torch versions, I can update the dependency to:

"torch>=2.3.1",


and remove the upper bound as suggested.
Please let me know if you'd like me to make this change — I can update the PR and sign off accordingly.

gabe-l-hart: Yes, let's go ahead and simply remove the upper bound. If CI passes, this will be a good additional piece of flexibility for users and will enable forward-compatible CVE fixes.

pasalubong: @rpancham Have you had a chance to circle back on this change?

